PR #404, opened 2015-05-02
Gbe 400

jonkiparsky:
Since I have to have unit tests before I tackle views and models, I'm going to ask that these be merged now, before they get tangled up with the django-cms merge.


bethlakshmi:
Sorry to take so long...  looks good.

PR #411, opened 2015-05-27
Gbe 403

jonkiparsky:
According to the coverage tool, this gets us to 76%, for what that's worth. Still PLENTY to be done about testing - for example, this just tests that the functions run, it doesn't really test that they run correctly - but this is a pretty good first step.


bethlakshmi:
FYI - there were merge conflicts, but I had trouble identifying an actual problem - it seemed to be a formatting issue, so I merged with my github tool rather than through the web tool.

PR #420, opened 2015-06-18
Adding mailchimp integration. This is implemented as a templatetag. T…

jonkiparsky:
…wo things are necessary to make this work

1) in your local_settings, define MC_API_KEY, MC_API_URL, MC_API_USER, and MC_API_ID
2) load gbe_tags and use the tag {% mailchimp %} to include a standard mailchimp "super-slim subscribe" box

Note that if MC_API_KEY is not defined, django will assume that mailchimp is not set up, and simply not display the box.

### DIFF
--- a/expo/expo/settings.py
+++ b/expo/expo/settings.py
@@ -279,3 +279,8 @@ CMS_PERMISSION = True
 
 
 CMS_PLACEHOLDER_CONF = {}
+
+try: 
+   MC_API_KEY
+except:
+   MC_API_KEY = 'TEST'  # if not set, we won't try to hit mailchimp's API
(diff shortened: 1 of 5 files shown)


bethlakshmi:
Do you have a recommendation on what the MC_API_KEY, MC_API_URL, MC_API_USER, and MC_API_ID should be?  I'm assuming we don't have to install anything... just point it to the right place?

jonkiparsky:
You can set up a test account for your own use on mailchimp, that's easy and free. We'll get the burlesque expo info from Scratch.

burlexpo:
The Expo does not currently have an active Mailchimp account.  Do I need 
to set one up?

jonkiparsky:
I suppose you might as well, if we want to make use of this functionality. :)

PR #425, opened 2015-06-20
Basics of contact form

jonkiparsky:
This pull request gives us the minimal case of a contact form. USER_CONTACT_RECIPIENT_ADDRESSES controls the recipient list. 

Some caveats:
*I haven't done anything about styling of the contact form, this is just about the functionality.
* I'm concerned about the testability of this code, but addressing that means going back to the vagrant idea - setting up a local testing machine provisioned for testing stuff like sending email. That's a big job
* One improvement to this will be to log the use of this feature, since a form of this sort is obviously a spam magnet. Need to approve and merge the logging branch for this to happen. 
* Another improvement will be to throttle the use of this rather severely. For example, rejecting out of hand any attempt to send more than once in any five-minute period. 
* Also, might want to allow blacklisting by IP. (the need for this will depend on the quantity of spam received)
* This should really be handled by an AJAX call rather than posting back to the server. 
* etc, etc - as I say, this is the minimal version. Reviewers: if there's anything you need to see before you can call this done, please add comments to this pull request.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,5 @@ expo/expo/local_settings.py
 *.bettytest
 
 expo/db.sqlite3.cmsnew
+
+expo/src/pip-delete-this-directory.txt
--- a/expo/gbe/forms.py
+++ b/expo/gbe/forms.py
@@ -676,3 +676,13 @@ class ClassScheduleForm(forms.ModelForm):
                   ]
         help_texts = classbid_help_texts
         labels = classbid_labels
+
+class ContactForm(forms.Form):
+    '''Form for managing user contacts. Notice that there 
+    are no models associated with this form.
+    '''
+    name = forms.CharField(required=True)
+    email = forms.EmailField(required=True)
+    subject = forms.CharField(required=True)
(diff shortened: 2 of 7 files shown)


burlexpo:
On 6/20/15 4:54 PM, Jon Kiparsky wrote:
>
> This pull request gives us the minimal case of a contact form. 
> USER_CONTACT_RECIPIENT_ADDRESSES controls the recipient list.
>
{snip}
>
>   * One improvement to this will be to log the use of this feature,
>     since a form of this sort is obviously a spam magnet. Need to
>     approve and merge the logging branch for this to happen.
>
Can we add a CAPTCHA functionality? Or a similar proof that you're a 
human (I've seen a CAPTCHA alternative that requires solving a simple 
math problem).

bethlakshmi:
Got it.
Template conflict between the mailchip and the contact form.
FYI - we should muck with the layout more - in the 3 column layout this form doesn't display so well - the window has to be pretty big or the form gets cut off.  I figure Scratch may have other style stuff, so figured I wouldn't fret about it.

PR #477, opened 2015-07-26
Gbe 476 - problems with contact form

jonkiparsky:
Contact form works now.


bethlakshmi:
I just approved and pulled and now it's not working...
The only change I did locally was to turn Debug to True

PR #482, opened 2015-07-26
GBE 472

bethlakshmi:
Edits from Scratch
Minor changes to formatting
Found a way to make class proposal a plugin so Scratch can change at will.

### DIFF
--- a/expo/gbe/cms_plugins.py
+++ b/expo/gbe/cms_plugins.py
@@ -2,8 +2,25 @@ from cms.plugin_base import CMSPluginBase
 from cms.plugin_pool import plugin_pool
 from django.utils.translation import ugettext as _
 from django.template import loader
-from gbe.forms import ContactForm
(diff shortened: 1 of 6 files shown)


expo/gbe/cms_plugins.py
```diff
@@ -2,8 +2,25 @@
 from cms.plugin_pool import plugin_pool
 from django.utils.translation import ugettext as _
 from django.template import loader
-from gbe.forms import ContactForm
+from gbe.forms import *
```
jonkiparsky: I would actually like to use explicit imports in general, in this style:
'''
from gbe.forms import (
    ContactForm, 
    FooForm, 
    BarForm,
    )
'''

jonkiparsky:
Otherwise, looks good.

bethlakshmi:
Updated with change from Jon.

PR #483, opened 2015-07-26
GBE-385: Require user ID info to be unique

jonkiparsky:
This does not force purchase email to be unique (and does not prompt for purchase email to be changed when user email changes). If there are any more requirements about purchase email, we should cover them on a separate ticket. 

This does expose purchase email, which seems necessary.


bethlakshmi:
I see and will look when I have more than 2 seconds...

bethlakshmi:
Uh-oh, this has a problem - 
- when I go to "/update_profile" - I can update everything EXCEPT my email - if I set my email to a different value, everything acts OK, but the change doesn't stick.

Also - of note, there is a secondary thread of updating this data - 
- /profile/manage - is a listing of users that the registrar can use
- by clicking on any user and editing, they get to "/profile/admin/1"
- this lets the registrar update the purchase email.

Right now that breaks completely, the registrar can't update anything on this page, because the email reads as a duplicate.

jonkiparsky:
>
>
>    - when I go to "/update_profile" - I can update everything EXCEPT my
>    email - if I set my email to a different value, everything acts OK, but the
>    change doesn't stick.
>
>
Are you talking about the part where your PURCHASE email doesn't change,
and that's shown in the center pane of your home page, so you think it's
supposed to be the thing that changed? Yeah, that got me too. :)
(this is why I exposed the purchase_email field... )

Looking at the admin path now

jonkiparsky:
Fixed profile/manage.

bethlakshmi:
I'm good now if you're good.  I'm the last changer, so the honor is yours...

bethlakshmi:
Sigh... that was not the branch I meant...

jonkiparsky:
Jon Kiparsky's TicketsNo tickets have been purchased, if jon@kiparsky.net
is not the email you used when purchasing tickets, please contact us for an
update.

This always shows the purchaser_email. This is going to cause confusion
when someone changes their email address and sees the old address on their
/gbe page. Scratch will hate this later, might as well deal with it now.

bethlakshmi:
Nope... it was the user email field.

See note in checkin - we tell them the purchase email, but don't let them
update it. It's messed up, but I can't lay this one on Marcus.  Our
special, special users have a tendancy to buy tickets using other people's
pay pal addresses - so the separation of "email I used to buy stuff" and
"email I want for this account and all GBE communication" of necessity gets
split and then it gets weirder down that rabbit hole...

-B

jonkiparsky:
> Nope... it was the user email field.

Well, I've done nothing to fix that in update_profile, since it's been working fine for me. So if this was a problem for you, it's probably still a problem. (I've only fixed the admin flow, not the user flow)
Can you retest and send me a more precise breaking case, if you find one?

jonkiparsky:
@bethlakshmi Where do we stand on this? You raised an issue that I've been unable to replicate. Can you give me a better example?

PR #486, opened 2015-07-31
Gbe 443

jonkiparsky:
This should be good to go for Stage 1 and Stage 2. 

First: review the code carefully. Let me know if you have any concerns. 

Stage 1: 
1. take a copy of the production database, and a backup of the test db just to be safe. 
1. Load the prod copy onto test server
1. Sync db. Should work. 
1. Test current state. Nothing should work different except tickets will no longer be showing. (this is a hard-coded change)

Stage 2:
1. Create two new Conference objects, one for 2015 and one for 2016
1. in the shell, or in mysql if you prefer, point each biddable and each Event and each Ticketing Item at 2015.
1. Test current state. Nothing should be very different. 

Touch base with me when you hit this point, or if you run into any issues up to this point.


bethlakshmi:
I see it!  Will read and start pulling tonight before doing stage 1.

bethlakshmi:
I have a few concerns, but I'm willing to go forward and fix them and submit my own pull requests if I have to... they are:

- what is up with gbe/models.py - line 779?
   - it doesn’t do anything- I don't see how this changes what it means to be visible, so I don't see how 2015 bids get hidden, and I thought that was part of the aspect of this change.   If it isn't - this is a BLOCKER to going live.

- I don't understand the ticketing change in gbe/models.py - line 947 - by doing this, the events won't show tickets in their description.  Since tickets are manually assigned to events, I don't see the point of this.  We are hiding tickets from being displayed with events for no good reason.  I don't see this as a blocker, merely an annoyance we will wish we hadn't done, when we fix it later.

- I assume that expo/gbe/ticketing_idd_interface.py line 134 is a placeholder - it works for now, because it hides old tickets (however hideously) but should be removed when tickets are labeled per conference.

Mostly I'm confused on 779 - if we are hiding old conference bids, we should be hiding them.  This does not hide them.

bethlakshmi:
Summary - I'm moving forward with migrations - this gives me enough to work with that in light of deadlines, the net benefit is good.

jonkiparsky:
What you're seeing is where I took out - as promised - the parts of the
code that went beyond simply changing the model.
The reason that particular change had to go is that "is_current" depended
on the existence of the conference field, which would be irritating to you
in stage one and stage two.
Stage three is where we make things invisible - remember?

bethlakshmi:
Aha. OK.  Cool.  So... when is Stage 3?  Is that something you can commit tonight or something I'm waiting until Sunday on?  Mostly my confusion is that before we can go live, I need to hide bids from previous years.

Stage 1 is done.
Stage 2 step 1 is done
Moving onto Stage 2, step 2 - I assume it is actually Biddable, Event, TicketItem, and ClassProposal.

jonkiparsky:
Yes, ClassProposals also. 

I'm going to try to get hiding committed tonight. 

From there, creating will be the next step.

bethlakshmi:
Awesome!  

Stage 2 is completely done.
All looks well.

It seems OK to pull the DB back from test to live in prep for Stage 3?
Right?

jonkiparsky:
Sure. Stage 3 will involve a model change, of course, so we should probably have a copy of the current state lying around.

bethlakshmi:
Another one?  I thought the point of stage 1 and 2 was to finish it... OK... I will grab a dump, but if there's a change, no reason to reimport to live.

jonkiparsky:
Yes, we'll now have to make conference a required field. I suppose it's not strictly necessary, but it seems like a good step to me at least.

bethlakshmi:
Nope, fair enough.
Forgot that, I don't *think* it's actually a DB level change, I think django handles that in the ORM, at least with MySQL.

I didn't realize you could make that change yet, I figured that had to fit with the creation stuff in Stage 4, since it means new bids have to have the assignment.

Although.. if you wanted to set a default for it - the id of the GBE 2016 conference is currently 2.

jonkiparsky:
I think there's a DB constraint imposed by is_blank=False. Could be wrong.
Anyway, we do have a bit more work coming, never hurts to have a copy lying
around.

bethlakshmi:
OK.
Standing by when you're ready...  I have a local fresh db ready to play a bit.

PR #509, opened 2015-08-25
GBE-508: PEP8 cleanup of scheduler.models and scheduler.forms

jonkiparsky:
Inserting a bit of sanity in the scheduler


jonkiparsky:
@bethlakshmi Merged your changes into mine. Should be clean and good to go

PR #511, opened 2015-08-28
GBE 459

bethlakshmi:
Commented out failing tests.
Bit of pep8
model pep8, too, just for fun.


jonkiparsky:
@bethlakshmi - Looks good. There's definitely room for improvement in the codebase, and that will improve these tests but for now this is great.

PR #514, opened 2015-08-29
Gbe 513: fixed tab order on login page

jonkiparsky:


jonkiparsky:
@bethlakshmi, @BenignCremator Pretty simple change, but it fixes a minor annoyance.

PR #517, opened 2015-09-06
GBE 471: Make ticket list hierarchial and thus easier to use

bethlakshmi:
Finally!
- serious rearchitecture of tests to make factories reusable across applications
- serious rework of ticketing with DB changes.
- includes every part of conference connection that I could think of
- includes usability rework based on my experience last year


bethlakshmi:
Sigh.  Sorry for pull request drama.  I did something weird to Master last night... I believe I've recovered it.

expo/tests/gbe/test_bid_class.py
```diff
@@ -4,10 +4,13 @@
...
 from django.test import Client
 from gbe.views import bid_class
-import factories
-from functions import (
-    location,
-)
+import mock
+import gbe.ticketing_idd_interface 
+from tests.factories import gbe_factories as factories
+from tests.functions.gbe_functions import (login_as,
```
jonkiparsky: Unused imports

expo/tests/gbe/test_bios_teachers.py
```diff
@@ -6,7 +6,14 @@
 from django.test.client import RequestFactory
 from django.test import Client
 from gbe.views import bios_teachers
-import factories
+import mock
+from django.contrib.auth.models import Group
```
jonkiparsky: unused import

expo/tests/gbe/test_bid_class.py
jonkiparsky: Also, line break after paren to avoid pep-8 problems and make the style consistent.

jonkiparsky: And in general, please don't undo code cleanup!

jonkiparsky:
Need to clean up all of the copy-paste damage before I can merge this.

jonkiparsky:
Thinking about the high-level idea here, the change to the test structure seems like a bad reaction to our bad design decisions. Probably an unavoidable bad reaction, but a pretty bad one all the same. 
The reason we're having trouble with our tests is because our apps are not isolated - we've designed this thing badly. This reaction is bad because it further locks us into this bad design. Our tests are no longer connected to our apps - this is a terrible place to be, design-wise.

I'm pretty sure we're going to regret this in about a year's time - this might be the move that forces us to scrap all of this and rewrite. It's certainly the move that prevents us ever hitting white-label or otherwise sharing this code: there won't be any point in anyone else trying to use this code until we undo this, unless they want to take the trouble to undo this themselves. 

Willing to work towards a merge, if we have to, but I think we should look at this really carefully.

expo/tests/factories/gbe_factories.py
```diff
@@ -12,7 +12,7 @@ class Meta:
         model = conf.Conference
 
     conference_name = factory.Sequence(lambda n: "Test Conference %d" % n)
-    conference_slug = factory.Sequence(lambda n: u"test_conf %d" % n)
+    conference_slug = factory.Sequence(lambda n: u"test_conf_%d" % n)
```
jonkiparsky: Good call

jonkiparsky:
> All merge/paste errors fixed

???

I still see 
`
from tests.functions.gbe_functions import (login_as,
                                           is_login_page,
                                           is_profile_update_page,
                                           location)
 `
at the top of every test case. This is new, it's showing as added code in the diff. Why are we adding unused imports? (in most cases, these imports are not used)
I'm also pretty sure we're not mocking in all of these test cases, but mock is imported into all of them. 

I'm a little concerned about this because I've already cleaned this stuff up once before, and I really don't relish the thought of having to do it again.

bethlakshmi:
When I originally did a refactor of the structure of this code, I hit test execution errors all over the place - to the point where I couldn't understand how it was running in the first place - there were quite a lot of missing dependencies and also extra imports - to the point where I couldn't see that I was making MORE of a mess (in approximately 8/2/2015 code) by consistently putting in the same template of imports - every single file that I changed came from an error raised when I ran "./manage.py test".

Then I waited on the merge to get back from Poland - because I didn't want to induce a database change to master and when I got back a ton of clean up had been done.

I'm glad to spend some time tonight on cleaning up the last import of gbe.functions.

But I'd really like to call that the end of this - at this point the time to fix the code that was broken is about 1/4 of the time I've spent trying to get this merge right - and that strikes me as WAY too much process for a 3 person team.

I agree on the design interdependancies - but I think that trying to stick that as a blocking point of this merge is the same thing as cutting off one's nose to spite one's face.

bethlakshmi:
OK - ready to try again @jonkiparsky  - I went through the contents of tests/gbe/* specifically and I think I got all of them AND a few extra redundant imports. All tests pass, directory is pep8 error free.

expo/tests/factories/gbe_factories.py
bethlakshmi: Thanks.  It broke in the test case I was writing in ticketing... and then I verified that slugs don't like whitespace. :)

jonkiparsky:
Unfortunately, I'm getting test failures and errors. 

> FAILED (errors=2, failures=1)

The errors are in test_edit_event.py, both due to 
> DoesNotExist: LocationItem matching query does not exist.

The cases are test_edit_event_get_succeed and test_edit_event_submit_succeed

The failure is:
`
======================================================================
FAIL: add event post succeeds, user has proper privileges
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/jpk/nonlt/code/GBE2/expo/tests/scheduler/test_add_event.py", line 86, in test_add_event_submit_succeed
    self.assertEqual(response.status_code, 200)
AssertionError: 302 != 200
`

bethlakshmi:
I absolutely can't make these reoccur in my environment.

I have a couple of thoughts:
1 - do you have any residual *.pyc files?  I had errors exactly like these when I was working on the scheduler unit tests.

2 - is there a reason to believe that:

   fixtures = ['tests/fixtures/rooms.json'] 

Would work in one environment and not another?

And do you have the copy of the json file that master has?

Both problems could occur if you don't have a working, loaded fixture, since both could be coming from  code querying for LocationItems that are in the fixture (and if the fixture isn't loaded, the tests won't work).  The location change worked on my system with no changes needed to settings.py, but if there was anything different in your local settings.py (eg, explicitly set fixture directories?) then it could have blocked the movement of files.

bethlakshmi:
Did you run the tests before the approval of the last commit you approved - that's where I did updates to scheduler that created these test cases.

bethlakshmi:
... and are there any other copies of rooms.json in your environment - I'm given to understand that duplicate copies of the fixture in multiple places can cause it to load the first available.  I had thought that the path provided made it explicit enough - but there are always randomly difficult options.

bethlakshmi:
And yes... I know ... vagrant... Believe it or not, it was high on my list for this week, if I could just get this fix in...

jonkiparsky:
Cleaning up my .pyc fixed it. Merging.

useful alias: 
`alias rmpyc='find . -name "*.pyc" | xargs rm'`

(also: `alias pepdiff='git diff --name-only master | grep ".py" |grep -v "mc/migrations" | xargs pep8'`)

jonkiparsky:
Crap. I was wrong, I'm still getting the failures. We'll talk about it in the standup.

PR #520, opened 2015-09-07
Gbe 498: Show historical bids and allow users to "clone" them

jonkiparsky:
Combined tickets 497 and 498. Cloning works for Acts, Classes, and Vendors.


jonkiparsky:
I've merged recent changes to master into this branch, and made some necessary edits. Should be an easy merge now.

bethlakshmi:
With the historical flag off, when I create 2 new act drafts, I get a panel on my landing page that says:

 new act - Not submitted

    View/edit this act bid

    **Propose a new act**

act to accept - Not submitted

    View/edit this act bid

    **Propose a new act**

Proposing a new act should only happen at the bottom of the panel, it's not per-act operation.  That's how it worked pre-clone.

-- Classes have the same problem.
-- Vendors are fine.

bethlakshmi:
Other than that... can you also wrap a div around the text:

    We are only showing bids for the current conference. If you have proposed acts, classes, or vendors in previous years, click here

And give it a class.

And around the:

   Showing Historical Bids

  These bids are not currently available for consideration. Click the "clone" button to re-submit an act, class, or vendor bid in the current conference.

  View current bids

And is there a way we could mark the whole side bar when it's in historical bids mode (maybe put in a class on the big box called "history") so that's super obvious that you're in historical mode.

I think the feature works great, I'm just trying to figure out ways to make it less subtle.  I actually missed the text the switchover for the first hour of messing with it.  If I'm that clueless, people who aren't me will likely be worse.

jonkiparsky:
@bethlakshmi Made fixes, please re-review

bethlakshmi:
I still saw it in classes.
And when the new sidebar class came in, the CSS stopped coloring the act, performance, vendor, etc. divs since they were relational.

I just did a check in with a bunch of visual only updates.  

Then I found one bug - a troupe member shouldn't be able to clone an act unless they are the contact for the act.  They can, however, view the act.  So I removed the clone link when this is the case.

@jonkiparsky - please review my last 2 checkins.  I'm ready to approve - so if you are, please make the final approval.

PR #524, opened 2015-09-11
Three bad imports.

jonkiparsky:
@bethlakshmi should be an easy review


bethlakshmi:
smacks head.
There was a very long 2-3 hours of Betty vs. Json vs. unit testing in this area.
I blame that.

Issue #535, opened 2015-09-25
Add some useful aliases + coverage tool


jonkiparsky:
"coverage" is a handy tool which analyzes your test coverage and produces nice reporting. I'm adding it to the stack so we can get a sense of where our coverage is weak. 
It's not exactly intuitive to run it, so I've adapted an alias that we use at MassChallenge. 
`$ runcoverage`

will run all of the tests and produce a handy report in the /vagrant/htmlcov directory. You can find your way to this directory, and bookmark the index.html file that you have there. We can have a tech huddle some time to talk about what to do with this information, but for now it's useful for me to have it around.
(note: it was written by Ned Batchelder, who hosts the Boston Python meetup and is a really nice guy)

Other aliases you might like:
rmpyc: recursively descend and remove all .pyc files. Useful for cleaning up deadwood, for example, if a source file is removed. 
pepdiff: report pep8 problems just on the files that are changed in the current branch (ie, different from master). This way you can focus on the files you've worked on. :)

This should be an easy merge, doesn't actually touch the codebase so no tests required. (but please do a vagrant destroy/ vagrant up cycle to verify that I haven't broken the build by including coverage in requirements.txt)

PR #536, opened 2015-09-25
Gbe 535: Add coverage tool plus some useful aliases

jonkiparsky:
See text of issue #535 for some details. Should be an easy PR to accept, since none of this actually touches the codebase. (but as I say in the ticket, should probably do a vagrant destroy/vagrant up cycle just for form's sake)


jonkiparsky:
Oops, I think I got some extra stuff here! Closing, will try again.

PR #537, opened 2015-09-25
GBE-535: Adding coverage tool and some aliases

jonkiparsky:
Some nice extras. The way I've handled the aliases (copy them over to the vagrant user's home directory and add a line to the .bashrc to get them loaded) seems a bit crude, but it works. 

See ticket #535 for some details about what this stuff is about.


jonkiparsky:
I've merged in the current master and re-run the tests. All still pass.

bethlakshmi:
looked good.
Merged, set up my vagrant again, made sure server starts up, and then ran unit tests.

PR #541, opened 2015-10-17
GBE-128: Provide FriendlyURLInput widget for URLFields. 

jonkiparsky:
Standard widget just specifies an html input with type=url. This doesn't work because different browsers validate URL input in different ways. I've used a pattern that recognizes a url as an optional leading "http://" or "https://", followed by at least two strings separated by periods, with an optional document path specifier (ie, /index.html or /jon/index.html)
Some examples of acceptable urls are as follows:

http://foo.com  
http://foo.bar.com
http://foo.bar.com/~quux
https://foo.bar.com/~quux

I'm not entirely happy with the regex I came up with, but it works for all of my test cases. Breaking test cases would be a great incentive to improve it.


BenignCremator:
DL'd the code to a vbox and tested it.  The URL field input is nicer then it was before.  Considering this done for now, and merging.

PR #545, opened 2015-10-27
Gbe 516: Event list should have filter for conference

jonkiparsky:


jonkiparsky:
@bethlakshmi - I've done a merge with current master (including GBE-515), and this passes all tests on my machine. I hope this makes your life a little easier.

bethlakshmi:
I can't break it! Whoohoo!

PR #561, opened 2015-11-15
Gbe 503: add conference drop-down to bid review page

jonkiparsky:
Added dropdown. Note that this still links to active review pages for old bids, that will be fixed in GBE-504

### DIFF
--- a/expo/gbe/models.py
+++ b/expo/gbe/models.py
@@ -53,6 +53,11 @@ class Conference(models.Model):
         except cls.DoesNotExist:
             return cls.current_conf()
 
+    @classmethod
+    def all_slugs(cls):
+         return cls.objects.order_by('-accepting_bids').values_list(
+             'conference_slug', flat=True)
+
     class Meta:
         verbose_name="conference"
         verbose_name_plural="conferences"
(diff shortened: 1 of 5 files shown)


BenignCremator:
Have checked and tested the described functionality.  What is provided works on the website.  Checked selecting both conferences on review acts, review proposals, and review classes, and conference selection works there.  Does not occur to be included in review tech information, but that also seems to be out of scope.  Also ran the related tests for test_review_act (FAILED), test_review_act_list (passed eventually), test_review_class (passed), test_review_class_list (passed), and test_review_proposal_list (passed).  The failed test failed with error: 'IntegrityError: (1048, "Column 'conference_id' cannot be null")', which I am assuming means that this test was not updated with the rest of the code, and that added a conference id will cause it to pass.  However, a non-functional test is part of the pull request and functioning code, so I am not merging the pull request until the test is updated, or someone with more cred then I do decides that this is ok. either as an exception, or by indicating that the failing test is not as relevant as I assume.

jonkiparsky:
Interesting - not getting this error when I run all the tests, so you've found a weird dependency. I was able to resolve it by changing from a RelatedFactory to a SubFactory, but why this makes a difference I'm not sure. 
Good catch.

BenignCremator:
Retested.  Nothing broke, and all tests are working properly.  Merging.

PR #570, opened 2015-11-17
GBE 564

### LINKED ISSUE #564: Make Admin easier to use
I'm pondering just doing this tonight.

Now that we have the conference linked in to events, tickets and bids, we could really use the Admin pages to have that as a filter.  These are our back up for WTF type problems in the admin screens and also a way that lets us sometimes avoid writing specialty reports for Scratch.

I say, let's just have one ticket and create a filter per conference every where is a dependant class.

bethlakshmi:
closes #564


jonkiparsky:
Please add some more details, either here or on issue #564, about what exactly I'm looking for when I review this.

PR #571, opened 2015-11-17
Gbe 556: Fix filter for all events

jonkiparsky:
While I was at it, I cleaned up some ugly dependencies in scheduler/views.py and unified the permission functions (which were previously defined in two places)


BenignCremator:
Have tested code, and it works as described.  However, I am getting more failed tests.  They are generally of the 'conference id can not be null' type when some tests are run singularly.  Running all tests works just fine.  Merging in from master to see if that fixes the problem.

BenignCremator:
That did it.  Now all tests pass when run singularly.  Merging.

PR #574, opened 2015-11-19
Gbe 554: Act coordinator act acceptance displays old conference's shows

jonkiparsky:
Functionality works.


bethlakshmi:
Thanks Jon!

Issue #552, opened 2015-11-11
Vanishing/reappearing edit bug


burlexpo:
Again.

I haven't touched it.  I went to browse the page and there's no data there -- just a menu item.

I am *extremely* nervous about stuff randomly disappearing in general, more so when we don't yet have a back-up scheme in place. (Do we?)

burlexpo:
And it has vanished from the test site too.

BenignCremator:
On 11/11/2015 09:05 AM, Scratch wrote:

> "Roll of Winners" page has vanished

That is not good.  Is the 'Roll of Winners' hosted as a static page or 
off-site page?  It seems to be back now, did someone fix it, or is there 
something happening intermittently?

> I am /extremely/ nervous about stuff randomly disappearing in general,
> more so when we don't yet have a back-up scheme in place. (Do we?)

Backups are occurring to my personal server.  They seem to be running 
ok, but I am still in the process of monitoring their progress. 
However, what we do not have is a full recovery plan in case we need to 
use the backups.  Also, I added audio to the backup scripts, as I had 
missed that the first time around.

Hunter

jonkiparsky:
I am also seeing the page on the site at present (the cropping on Kiki Allure's shot needs attention, but the page is there). I haven't been anywhere near the live site in quite a while, all of my work has been at the repo end. @bethlakshmi, have you been on the server? If not, this is a little bit mysterious. 

@burlexpo  - can you give more detail about what "vanished" means? Was the menu item present? If so, what happened when you went to it? If not, did you try going directly to the url, and what did you get when you did?

burlexpo:
Ok -- more details.  What seems to have "vanished" is my ability to edit said page (which I'm not sure I ever had).  What happens is, when I'm *not* logged in, I can see the page.  When I *am* logged in (and therefore in edit mode on the CMS, since there doesn't seem to be a way for me to be logged into the site but not logged into the CMS), the menu link is there, but the page has no content (to me).

Does this help?

jonkiparsky:
Very much so. I'm looking at the page now, and when I'm signed in, I can see content on the page, but when I select "edit", the content goes away. I don't know if this is exactly what you're saying, but it seems like it's pretty close. 

I'm pretty sure this is a CMS configuration issue. @bethlakshmi has a lot more experience in this area than I do, so I'm hoping she'll chime in on it, but I can try to understand it on my local copy if that'll be useful.

burlexpo:
[...]

PR #579, opened 2015-11-24
GBE 562

bethlakshmi:


bethlakshmi:
Sorry  - meant:
https://github.com/bethlakshmi/GBE2/issues/552

And read the middle "5" as a 6.  Not sure why.

The fix in the end was:
- stop the site from flipping between "en" and "en-us" by removing "en" as an option.  This settings.py update fixes that.

PR #580, opened 2015-11-24
GBE 552:  Remove "en" to fix vanishing & reappearing Edit bug in Django CMS

### LINKED ISSUE #552: Vanishing/reappearing edit bug

bethlakshmi:
Closes #552.

This time as the right issue number.

### DIFF
--- a/expo/expo/settings.py
+++ b/expo/expo/settings.py
@@ -95,28 +95,28 @@ INSTALLED_APPS = (
     'south',  # Only needed for Django < 1.7
     'sekizai',  # for javascript and css management
     'djangocms_admin_style',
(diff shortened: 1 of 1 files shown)


BenignCremator:
Ran server and tested CMS.  en-us is the only language option available, can not create a en page, or any non en-us pages.  Also grepped more recent db dump for "en" and grep -v'ed "en-us" and got 0 results, so no en pages in the db.  This all looks good, merging.

PR #582, opened 2015-11-29
GBE 549

### LINKED ISSUE #549: Act Tech Info Form looks bad
The preview of act tech info submitted information looks pretty bad - it needs a header and better layout.

bethlakshmi:
Closes #549

Sorry not to have a new unit test here, but we have no mocking yet on test_edit_act_techinfo, and it seemed crazy to spend time on a test when there’s no change to business logic.  Just a 2 line change to visual display.

### DIFF
--- a/expo/gbe/static/styles/base.css
+++ b/expo/gbe/static/styles/base.css
@@ -747,3 +747,10 @@ td.form_field > ul > li
 {
   list-style-type: none;
 }
+
+.formtitle
+{
+  font-family: "TwentiethCenturyPoster";
+  font-size: 20px;
+  color: black;
+}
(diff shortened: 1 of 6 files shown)


bethlakshmi:
Scratch the comment about no unit test ready.  I made one.  It was bugging me and worth it to at least have one.  The test DOES include the test of the code I added.

jonkiparsky:
I like the style changes, and all of that looks good to merge. However, I note that we lose the header information after we submit the form. 
(To reproduce: create an act, mark it as submitted, and accept it into some show. Fill in some fields in the tech info. Save changes. Note that information about act - Title, Description, etc - is now blank)

jonkiparsky:
Hm. Looking at master, I see that this is happening on the original, so I suppose it's an existing bug. 

Merging this and logging that issue.

bethlakshmi:
Wow that's a good catch, though

PR #590, opened 2015-12-04
GBE 589

### LINKED ISSUE #589: Request to change Submit Tech Info form
After discussing the Submit Tech Info form with other stakeholders/tech crew, we have decided upon some changes we would like to have made.  They are:
1. Move the Cue Sheet (both form and example) to the bottom of the page, below everything else
2. Reword the Cue Sheet instructions, first paragraph, add between 3rd and 4th sentence: "Please keep cue descriptions and instruction as concise as possible, and give times for triggers if possible."
3. Add a Text Block, same format as "Color(s) of your costume :", between 'Color(s) of your costume :" and "Audio Info", labeled "Please Describe your Specific Lighting Needs:"
4. Remove the Cue Sheet (both form and example) from the form for the Newcomers Showcase. 

I would like 1, 2, and 3 for the Dec 5th Milestone, 4 can occur later.  Number 3 can slide, if we end up pressed for time, but is higher priority then 4.

bethlakshmi:
Closes #589

I got all four things.
Basic integration testing done.
Give it a visual look, I took some liberties with the cue text for readability.
@BenignCremator  - wouldn’t mind your thought on the text & flow.
@jonkiparsky - sorry there is not  more unit testing here. I don’t think I changed it at the level we’ve typically been testing, but the short time span is tough.  Only the last checkin is something that is seriously unit test worthy.

### DIFF
--- a/expo/gbe/migrations/0002_auto__add_field_lightinginfo_specific_needs.py
+++ b/expo/gbe/migrations/0002_auto__add_field_lightinginfo_specific_needs.py
@@ -1,0 +1,337 @@
+# -*- coding: utf-8 -*-
+from south.utils import datetime_utils as datetime
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+        # Adding field 'LightingInfo.specific_needs'
+        db.add_column(u'gbe_lightinginfo', 'specific_needs',
+                      self.gf('django.db.models.fields.TextField')(default='', blank=True),
+                      keep_default=False)
+
+
+    def backwards(self, orm):
+        # Deleting field 'LightingInfo.specific_needs'
+        db.delete_column(u'gbe_lightinginfo', 'specific_needs')
+
+
+    models = {
+        u'auth.group': {
+            'Meta': {'object_name': 'Group'},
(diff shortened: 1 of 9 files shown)


BenignCremator:
Loaded the new code and a DB, and tested all four bullet points.  Number 4 was still there, I assume because the information I was looking at was from last year, not this year.  Could not submit an act myself due to not having a BPT emulator, and conference selection not being complete.  However, since the requirement for merging was 1, 2, and 3, still merging since they all work, and the problems I encountered were not relevant to the new code.  Suggest keeping ticket open until further tests can be completed.

bethlakshmi:
There is one part not implemented - the Show switch for cue sheets is not easy to find.

<domain>/admin and then the GBE->Shows section has the switch for whether a cue sheet for the show should be 'Theater', 'Alternate' (ie, vendor room), or 'None'.  By default the show gets "Theater" - If you go to last year's new comer's show case you can give it "none" and get none.  

Design limitations:
- changing the cue_sheet doesn't change the cues.  
- 3 cues are made for any act always.  If the cue doesn't have instructions, the code won't display it on the reporting side... so if you watch the newcomer showcase act tech updates in the DB, you'll see 3 cues.  They are blank.   

I'll set up the new shows Scratch made in the DB... after the master update.

Issue #140, opened 2014-10-01
Run on Text Description on Vendor Display


burlexpo:
I created a vendor with a very long 'description of goods and services'.  If you take a look at http://gbeadmin.webfactional.com/deploy_stack/vendors/fashion_faire/ you can see the results.

Solution: put a hard limit on the number of characters that can be entered into the text box when the vendor bid is submitted.  I suggest 250 characters.

burlexpo:
I think we have a functional solution in place and this can be closed.  Can it?

jonkiparsky:
No, solution is not in place. Adding it now.

PR #607, opened 2015-12-12
GBE-140: truncate long descriptions to 250 characters on display

jonkiparsky:
Another trivial change.


jonkiparsky:
Closes #140

PR #631, opened 2015-12-22
GBE-630: Remove duplicate cases of worker in event details

### LINKED ISSUE #630: Staff Lead appears twice on page
Why? This may have something to do with the way I created the event(s)?

http://burlesque-expo.com/scheduler/details/403

screenshot 2015-12-20 20 21 56

bethlakshmi:
Closes #630.  

I would have rather done a distinct(‘role’, ‘_item’), but Django 1.6 with MySql doesn’t support this (it requires Postgres).

I tried a number of value() permutations but that flattens the query set into a dict, which breaks the expectations of the functional footprint.

This removes the duplicates from this specific thread.  The function is used other places, so one bonus is, it doesn’t change any other thread.

### DIFF
--- a/expo/scheduler/models.py
+++ b/expo/scheduler/models.py
@@ -574,7 +574,7 @@ class EventItem (models.Model):
                  Q(role__in=['Teacher',
                              'Panelist',
                              'Moderator',
-                             'Staff Lead']))).distinct().order_by('role')
+                             'Staff Lead']))).distinct().order_by('role', '_item')
         except:
             people = Worker.objects.filter(
                 allocations__event__eventitem=self.eventitem_id,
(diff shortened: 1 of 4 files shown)


jonkiparsky:
This is the sort of thing that we should write a test for - some test that would fail without this code, and passes with it.

bethlakshmi:
I'm never against unit tests but in this one, my rationale was that the only change is in the edge of the logic- the template.  We don't generally have template tests at this depth, and right now scheduler has absolutely NO unit tests.  Building out the unit testing here would both conflict with an area where @BenignCremator is working (scheduler unit tests) and also take a long time for a low value.

Instead, I'm putting in a ticket:
https://github.com/bethlakshmi/GBE2/issues/640

So we don't loose track of the issue and what the test case should be.

When there are more scheduler tests to build from, I'm glad to do it.

jonkiparsky:
Not merging this without a test, so I wrote the test. 
Closes #640

Issue #628, opened 2015-12-20
Contact Volunteers is broken


bethlakshmi:
http://localhost:8282/scheduler/contact_by_role/Volunteers
 Error = DoesNotExist at /scheduler/contact_by_role/Volunteers

This is unique to volunteers right now.

"High" because it doesn't work and there's no work around.

PR #636, opened 2015-12-27
Gbe 628: Error on contact volunteers

jonkiparsky:


jonkiparsky:
Closes #628

bethlakshmi:
Fix was good.
Unit tests broke - first because the permission check preceded these tests on master, so fixed merge conflict.
Then because the permission checking requires a profile, not just a user object.  So fixed that.
Now able to merge.

Issue #625, opened 2015-12-20
Act Tech review form - conference chooser needed.


bethlakshmi:
http://localhost:8282/reports/acttechinfo/view_summary

1 - put a conference chooser at the top (like bid reviews) and then have the list of shows at the top be the list of shows for the currently selected conference. Right now, they show the list of EVERY show and that's more than most users will care about most of the time.

Marked medium because the year reference of '15 on old shows is an OK work around in this area.

PR #639, opened 2015-12-27
Gbe 625: Add conference picker to act tech info review list

jonkiparsky:


jonkiparsky:
Closes #625

expo/gbe/report_views.py
```diff
@@ -173,28 +175,36 @@ def personal_schedule(request, profile_id='All'):
...
+            pass
 
     return render(request,
                   'gbe/report/act_tech_review.tmpl',
                   {'this_show': show,
                    'acts': acts,
-                   'all_shows': conf.Show.objects.all()})
+                   'all_shows': conf.Show.objects.filter(
+                       conference=conference),
+                   'conference_slugs': conference_slugs()},
```
bethlakshmi: When the user does the following:
1 - go to http://localhost:8282/reports
2 - select "Act Tech Info"
3 - pick an old conference

Then the conference picker defaults back to the current confernce.  If you send in the conference item, the picker will default show the currently selected conference.

bethlakshmi:
Not sure where to point to in the code, but:

When the user does the following:
1 - go to http://localhost:8282/reports
2 - select "Act Tech Info"
3 - pick an old conference
4 - pick a show from the old conference
5 - the show is shown 
problem = the conference list defaults back to an upcoming conference, and not the conference the user is looking at.

The context of what the user is doing should stay the same.  Given the similarity of shows from year to year in GBE, it would be easy to loose track.

bethlakshmi:
A last good thing would be that the conference picker on 
    http://localhost:8282/reports
has it's choice persisted to 
   http://localhost:8282/reports/acttechinfo/view_summary/

So that where ever the user has selected a conference, it stays consistent.

bethlakshmi:
That's all I have.  I don't mind making a ticket for later on the last comment (consistency between reports and act tech summary).  But the part where the show is chosen and then selector flips back to a default conference will make using this really frustrating.

jonkiparsky:
@bethlakshmi Done.

jonkiparsky:
@bethlakshmi this should do what you want.

bethlakshmi:
Sorry this took so long!!

PR #645, opened 2015-12-29
Gbe 555: Add filter to calendars

jonkiparsky:
This looks like it works, and I have one test that covers the simplest case, but should probably be tested pretty hard.


expo/scheduler/functions.py
```diff
@@ -391,6 +398,14 @@ def day_to_cal_time(day='Saturday', week=datetime(2015, 02, 19, tzinfo=pytz.time
     return cal_times
 
 
+def cal_times_for_conf(conference):
+    from gbe.functions import get_conference_days
+    days = get_conference_days(conference)
+    first_day = days[0].day
```
bethlakshmi: Does this assume that the first day in the get_conference_days call is the first day of the conference?  If so, then the get_conference_days should force a sort order.  It currently does not.

expo/scheduler/views.py
```diff
@@ -980,12 +977,16 @@ def calendar_view(request=None,
     # I want to rewrite this to get the first day of the event from the
     # DB or conf, and then calculate the date based on the next day after
     # the start of the event.  -  HH
-    if day is not None:
-        cal_times = day_to_cal_time(day,
-                                    week=datetime(2016,
-                                                  02,
-                                                  4,
-                                                  tzinfo=pytz.timezone('UTC')))
+    conf_slug = request.GET.get('conf', None)
```
bethlakshmi: Small nit - but could we use either 'conference' or 'conf_slug' for this name for the get.  We use either of those things for the conference slug marker and consistency is nice.

expo/scheduler/views.py
```diff
@@ -980,12 +977,16 @@ def calendar_view(request=None,
...
+    if conf_slug:
+        conf = get_conference_by_slug(conf_slug)
+    else:
+        conf = get_current_conference()
+
+    # ignore day parameter for now, day_to_cal_time ignores it
+
+    cal_times = cal_times_for_conf(conf)
```
bethlakshmi: BLOCKER - Something is broken here - when I enter:
http://localhost:8282/scheduler/Show/Sunday?conf=GBE2016
or 
http://localhost:8282/scheduler/Show/Sunday

I still get Saturday's data.  This really did work last year, I can see it working for last year's conference here:

http://burlesque-expo.com/scheduler/Show/Sunday

We can't call this ticket fixed without this.

bethlakshmi:
I think that's all my comments - I didn't look at the test code yet, but I suspect it will not be controversial.
@jonkiparsky  - the first two things were small nits, but the thing marked blocker is a blocker.  We can't just have a calendar for Saturday... we really need all days.

I don't mind changing from day of week to actual dates, if that makes this easier  for example, instead of
"Saturday?conf=GBE2016", It would be ok with me to do something like "2016-02-06?conf=GBE2016" if that makes the whole thing less irritating.

jonkiparsky:
@bethlakshmi This should clear up that problem.

Issue #463, opened 2015-07-24
New layout with Volunteer Scheduling for an Event is ugly


bethlakshmi:
it's not unusable, but it deserves a review.  The problem is on

/scheduler/edit/_any_event_class_/_event_id_
- where the event class is not of type Volunteer Opportunity

The Volunteer Management table is very long and as a result, it drifts way off the fixed width working area in the new layout.

It's 100% useable, the visual breakage isn't a blocker to the user.  But it's ugly and this is the placeholder to review when we have time.

It will benefit from the wise use of CSS, but I believe template work will be needed for this to be nice looking.

burlexpo:
@jonkiparsky I *think* this is the issue I was talking about...

jonkiparsky:
@burlexpo  I'll want to consult with you about this in more detail if I'm going to take this on. Probably a one-one thing. 
Can you modify the priority so it reflects your wishes?

bethlakshmi:
enough layout work happened recently that if we want it more useable, we need to redefine what actions we want.  The problems cited here are largely fixed on a laptop sized screen.

PR #647, opened 2016-01-03
GBE 650: Better Volunteer assignment

bethlakshmi:
Fairly big change - instead of the typical accept bid flow, a new thread is introduced.  To black box test:
- go to http://localhost:8282/volunteer/reviewlist
- use a Volunteer Coordinator
- click the “Assign” link
- get the new form

——————
General layout was worked out (literally) on the back of a napkin w. @burlexpo 

The flow has not changed at all - submission goes through existing code.

Includes some bug fixes - get_sched was commented out, changes to volunteer windows affected the accuracy of the availability listings.  I added some tests to cover the regressions and hopefully prevent in future.

I left the change state code in the review screen, but removed the horrible list of events.  This allows the volunteer coordinator to reject, wait list, etc. the volunteers.

### DIFF
--- a/expo/gbe/functions.py
+++ b/expo/gbe/functions.py
@@ -146,24 +146,3 @@ def available_volunteers(event_start_time):
         if event_start_time in window_range:
             windows.append(window)
     return conf.Volunteer.objects.filter(available_windows__in=windows)
-
-
-def get_events_and_windows(conference):
-    events = sEvent.objects.filter(
-        max_volunteer__gt=0,
-        eventitem__event__conference=conference
-        ).exclude(
-            eventitem__event__genericevent__type='Rehearsal Slot').order_by(
-                'starttime')
-    conf_windows = conference.windows()
-    volunteer_event_windows = []
-
-    for event in events:
-        volunteer_event_windows += [{
-            'event': event,
-            'window': conf_windows.filter(
-                day__day=event.starttime.date(),
-                start__lte=event.starttime.time(),
-                end__gt=event.starttime.time()).first()
-            }]
-    return volunteer_event_windows
--- a/expo/gbe/functions.py
+++ b/expo/gbe/functions.py
@@ -11,6 +11,7 @@ from gbetext import (
     class_options,
 )
 from gbe.duration import DateTimeRange
+from scheduler.models import Event as sEvent
 
 def validate_profile(request, require=False):
     '''
@@ -136,3 +137,24 @@ def available_volunteers(event_start_time):
             windows.append(window)
     return conf.Volunteer.objects.filter(available_windows__in=windows)
     
+def get_events_and_windows(conference):
+    events = sEvent.objects.filter(
+        max_volunteer__gt=0,
+        eventitem__event__conference=conference
+        ).exclude(
+            eventitem__event__genericevent__type=
+            'Rehearsal Slot').order_by('starttime')
+    conf_windows = conference.windows()
+    volunteer_event_windows = []
+
+    for event in events:        
+        volunteer_event_windows += [{
+            'event': event,
+            'window': conf_windows.filter(
+                day__day=event.starttime.date(),
+                start__lte=event.starttime.time(),
+                end__gt=event.starttime.time()).first()
+            }]
+
+    return volunteer_event_windows
+
--- a/expo/gbe/functions.py
+++ b/expo/gbe/functions.py
@@ -7,11 +7,12 @@ from django.core.exceptions import PermissionDenied
 from django.contrib.auth.models import User, Group
 from django.conf import settings
 from gbetext import (
-    event_options, 
+    event_options,
     class_options,
 )
 from gbe.duration import DateTimeRange
 from scheduler.models import Event as sEvent
+
 
 def validate_profile(request, require=False):
     '''
@@ -52,13 +53,13 @@ def validate_perms(request, perms, require=True):
         raise PermissionDenied
     return False               # or just return false if we're just checking
 
-'''
+
+def mail_to_group(subject, message, group_name):
+    '''
     Sends mail to a privilege group, designed for use by bid functions
     Will always send using default_from_email
-'''
-def mail_to_group(subject, message, group_name):
-
-    to_list = [user.email for user in 
+    '''
+    to_list = [user.email for user in
                User.objects.filter(groups__name=group_name)]
     if not settings.DEBUG:
         send_mail(subject, message, settings.DEFAULT_FROM_EMAIL, to_list)
@@ -66,30 +67,34 @@ def mail_to_group(subject, message, group_name):
 
 
 def send_user_contact_email(name, from_address, message):
-    subject = "EMAIL FROM GBE SITE USER %s" %name
+    subject = "EMAIL FROM GBE SITE USER %s" % name
     to_addresses = settings.USER_CONTACT_RECIPIENT_ADDRESSES
-    send_mail(subject, 
-              message, 
-              from_address, 
+    send_mail(subject,
+              message,
+              from_address,
               to_addresses)
     # TO DO: handle (log) possible exceptions
     # TO DO: log usage of this function
-    # TO DO: close the spam hole that this opens up. 
-              
+    # TO DO: close the spam hole that this opens up.
+
 
 def get_conf(biddable):
     conference = biddable.biddable_ptr.conference
     old_bid = conference.status == 'completed'
     return conference, old_bid
 
+
 def get_current_conference():
     return conf.Conference.current_conf()
-    
+
+
 def get_conference_by_slug(slug):
     return conf.Conference.by_slug(slug)
 
+
 def conference_list():
     return conf.Conference.objects.all()
+
 
 def get_events_list_by_type(event_type, conference):
     items = []
@@ -122,8 +127,9 @@ def get_events_list_by_type(event_type, conference):
         items = []
     return items
 
+
 def available_volunteers(event_start_time):
-    one_minute = timedelta(0,60)
+    one_minute = timedelta(0, 60)
     tz = pytz.utc
     event_start_time = event_start_time + one_minute
     windows = []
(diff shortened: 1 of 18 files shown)


jonkiparsky:
Getting a test failure:

======================================================================
FAIL: edit_troupe view, create flow
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/vagrant/expo/tests/gbe/test_edit_combo.py", line 41, in test_create_troupe
    self.assertTrue(self.troupe_string in response.content)
AssertionError: False is not true

expo/tests/gbe/test_edit_combo.py
```diff
@@ -16,7 +16,7 @@ class TestCreateCombo(TestCase):
     def setUp(self):
         self.factory = RequestFactory()
         self.client = Client()
-        self.troupe_string = 'a one-off group of performers working together'
+        self.troupe_string = 'a one-off group of performers working     together'
```
jonkiparsky: Why is this whitespace introduced?

expo/tests/gbe/test_review_volunteer.py
```diff
@@ -36,29 +36,3 @@ def test_review_volunteer_all_well(self):
         response = review_volunteer(request, volunteer.pk)
         nt.assert_equal(response.status_code, 200)
         nt.assert_true('Bid Information' in response.content)
-
-    def only_show_relevant_events(self):
```
jonkiparsky: Why is this test removed?

expo/tests/gbe/test_assign_volunteer.py
```diff
@@ -0,0 +1,275 @@
...
+        volunteer = VolunteerFactory.create()
+        request = self.factory.get('volunteer/assign/%d' % volunteer.pk)
+        request.user = self.privileged_user
+        request.session = {'cms_admin_site': 1}
+        login_as(request.user, self)
+        response = assign_volunteer(request, volunteer.pk)
+        nt.assert_equal(response.status_code, 200)
+        nt.assert_true('Assign Volunteer to Opportunities' in response.content)
+
+    def test_assign_volunteer_show_current_events(self):
```
jonkiparsky: A test should test one thing - this test tests many things. This is not a blocker, since we need to get this code merged, but this will need to be broken up into a suite of tests.

jonkiparsky:
Also, I'm not seeing any difference between the layout of scheduling pages - for example http://localhost:8282/scheduler/edit/GenericEvent/1371 - when I compare this to master. Specifically, both have the "Action" column hanging off the edge of the frame. What is the fix I'm looking for?

expo/gbe/functions.py
```diff
@@ -145,3 +146,24 @@ def available_volunteers(event_start_time):
         if event_start_time in window_range:
             windows.append(window)
     return conf.Volunteer.objects.filter(available_windows__in=windows)
+
+
+def get_events_and_windows(conference):
```
jonkiparsky: Surely a function which queries a table in the scheduler app should be under the scheduler app, no?

expo/gbe/models.py
```diff
@@ -314,21 +313,7 @@ def get_schedule(self):
         Returns schedule as a list of Scheduler.Events
         NOTE:  Things that haven't been booked with start times won't be here.
         '''
-        '''
-        from scheduler.models import Event as sEvent
```
jonkiparsky: This is good - we need to start getting the conference and the scheduler separated out, so this is a good step. (but see above)

expo/tests/gbe/test_edit_combo.py
bethlakshmi: It's coming from trying to fix the pep8 errors in either gbe_text or gbe_forms_text.

expo/tests/gbe/test_review_volunteer.py
bethlakshmi: 1 - it never ran - if you go to master, and remove this test, you'll see it doesn't change the test count at all when you run all the tests.
2 - it's unrunnable.  when I changed the function name form "only_show_relevant_events" to something that begins with the word "test_" it will run, and then gets lots of errors.  I don't think this test ever ran successfully - it was never executed at all so the errors in it where never obvious.
3 - when I refactored, it was obsolete.  The same test is now run in the assign volunteers testing, and now it works.

expo/tests/gbe/test_assign_volunteer.py
bethlakshmi: I'm glad to talk about test strategy, but I'll want a definition of what you mean when you say "one thing" vs. "many things".  It sounds like a simple and obvious statement, but the definition for how much a test case should test is actually not simple.

expo/gbe/functions.py
bethlakshmi: I disagree - this function also queries a table in the gbe app, so it should be in the gbe app, no?

My assertion - scheduler should not have to know about gbe since scheduler should work for other forms of conference logic.
I don't know that I can make the same assertion for gbe.

bethlakshmi:
The fix you are looking for is annoted in the comments of the pull request - there is absolutely no change to http://localhost:8282/scheduler/edit/GenericEvent/1371

This is entirely focused on changes to the bid acceptance process.

expo/tests/gbe/test_edit_combo.py
bethlakshmi: I can't replicate your error.  I did the following:
- go to the GBE-463 branch
- sync with github, just in case
- remove all *.pyc files in expo/tests/gbe
- remove all *.pyc files in expo/factories/gbe
- remove all *.pyc files in expo/

And I can't seem to find a way to make the test fail.  I'm willing to dump my vagrant box and try refreshing... but not tonight.

expo/tests/gbe/test_assign_volunteer.py
jonkiparsky: Simplest definition: one assert per test. This is too simplistic, of course, there can be reasonable reasons to use two or even three asserts to nail down a test, as above, but it's a good target. 
More nuanced definition: tests should be sufficiently focused that a failure of one thing does not mask other failures. For example, in this set of tests, if the string "Security/usher" does not appear in the response.content, then we have no information about any of the following asserts. This is a problem, since those conditions would give valuable information in diagnosing and fixing the problems, and also because it turns test fixing into a serial process - you can only fix one thing at a time, then you have to run the tests again and fix the next thing, when you might have got them all at once if you'd had the information.

expo/gbe/functions.py
jonkiparsky: Actually, this function doesn't query gbe_volunteerwindows. It takes as an argument an instance of Conference, and it calls a function on that object which returns some objects that have the characteristics that it's looking for. I agree that it violates the Law of Demeter, which is a problem that I was willing to put aside for another day, but the only model it directly queries is scheduler.Event. 

I only point this out because it's an example of the sort of thing that we're going to need to get under control if we're ever going to drag this thing out of the ditch.

jonkiparsky:
Then how is this a fix for #463?

> The problem is on
> /scheduler/edit/any_event_class/event_id
> where the event class is not of type Volunteer Opportunity

bethlakshmi:
Agreed.  It does not close this.

I know that I interviewed @burlexpo on "how exactly do you assign volunteers these days?" - in particular whether he looked at the volunteers and assigned them to things (the review process) or whether he looked at the schedules and then picked volunteers.  And he said he reviewed the volunteer and picked a schedule item.  So that's what I fixed.

But agreed that #463 is still open.

I'll gladly write a different ticket and update the pull request to keep the ticketing clean here, and we'll add to Thurs' conf call what do to about the actual #463.

expo/tests/gbe/test_assign_volunteer.py
bethlakshmi: I have trouble agreeing to "one assert per test" - I would have said "one set of conditions produces one definable result" and the number asserts should scope to what the defintion of the correct result is.

I'm definitely willing to agree that this is testing multiple conditions with different results and should therefore be broken down, to something like:
- if the volunteer is booked, the event is loaded as checked
- if the volunteer has an available window, they are shown as available
- if the volunteer is unavailable... 
- past conference events are not shown
- not-past conf. events are shown

etc... but that well may be more than one assertion, and keeping them as multiple different assertions gives me more information about what when wrong when a unit test fails.

I understand about the serial process problem, but trying a "1 assertion per test" seems to me like a lot of unnecessary repetitive setup.  Certainly it can be done well (a function call to do it) but I'm not sure that fully parellelizing in this direction is something I'd want to enfore or stick to.

expo/gbe/functions.py
bethlakshmi: Fair enough.  It's actually easy enough to move it right now if we think a decent guideline is:
- querying tables in other application models is higher on the list of things to avoid than accessing parts of objects from other application models since abstraction is better that way
- better to add more imports from scheduler functions 

The refactor is pretty easy and decidedly easier to do now than later.

bethlakshmi:
Argh.  OK - I made #650. 
This, when checked in, should close #650.
I tried to rename the branch, but I can't - I can make a new branch with the right label and pull a new request... or we can push ahead on this one... either, I don't want to kill this off until I know we actually have the comment threads in a good place....

jonkiparsky:
Don't worry about it - the branch only matters while it's active. After this is merged, the branch gets deleted.

jonkiparsky:
Figured out the test  failure - it was in a locally uncommitted file.

expo/tests/gbe/test_assign_volunteer.py
jonkiparsky: To me, it's clear that any sensible test can be checked with one or two asserts, and that any test that requires more than that is not a well-specified test. The fact that writing good tests scares you because of the "repetitive setup" is just you pointing out that our codebase is in lousy shape and we need to improve it - in this case, by factoring out the repetition and drying out the code. I agree with that, but it should be spurring us to improve it the problems, not to wallow in them. 

However, this does not need to be fixed here, as I said. I can merge this PR with this test in place, and we'll use it as a stalking horse for improving the code.

jonkiparsky:
Okay - looking at the thing that this is actually supposed to fix, I agree that it fixes it. I'm ready to merge it, apart from that stray function that needs to get moved.

expo/tests/gbe/test_assign_volunteer.py
bethlakshmi: Made #651 for remembering in the future.

bethlakshmi:
Updates have been made.

PR #649, opened 2016-01-05
GBE 581:  Landing page schedule fixed for multiple conferences.

### LINKED ISSUE #581: Personal schedule per conference
I believe we pulled personal schedules from display so as not to display old conference info.

We need to return them to being functional - I suggest:
- a personal schedule is available for any "current" and "upcoming" confernces, that we organize them in blocks per conference and then by day:
 - conference 1 (current first, upcoming second)
   - schedule item 1
   - schedule item 2 
 - conference 2
   - schedule item 1... etc.

~~Also - ASAP, we should pull the text saying that your personal schedule is appearing... it isn't...~~

bethlakshmi:
Closes #581
———————

Now testing is complete - it tests:
- that a current scheduled class is shown in regular mode
- that a past class is shown in historic mode
- that a current scheduled volunteer opportunity is shown in regular
mode
- that a past scheduled volunteer opportunity is shown in historic mode

And also vice, versa that things that are in regular mode are NOT in
historic mode and things in historic mode are NOT in regular mode.

Of note- getting the timestamps right is super nitpicky and I couldn’t
emulate it perfectly, but this should demonstrate that the timestamp is
not completely wrong.

Also - I chose not to test act assignment - it is one more logical
thread, but it was just one too many, and is not part of the code
changes that I affected in the update.

—————————

Functionality description:

1: visible on the landing page to any user:

http://localhost:8282/gbe

provided that the user has a role (resource allocation) in a scheduled event that is in an upcoming or current conference.

If there is more than one active conference (you can test this with our test data by setting “GBE 9” to be current - then each conference is shown ordered from past to future.

The conference’s name is now in the schedule to show which conference is which.

———

2 - if you select “View older bids” then the schedule is flipped and shows only past conferences.

——

A quick way to test this with relatively recent live data would be to look for users who have been booked into acts in this year’s shows.  Then they will have at least 1 schedule item (the show) and possibly 1 rehearsal if they have done their signup.

### DIFF
--- a/expo/tests/gbe/test_landing_page.py
+++ b/expo/tests/gbe/test_landing_page.py
@@ -18,80 +18,94 @@ class TestIndex(TestCase):
     '''Tests for index view'''
     def setUp(self):
         self.factory = RequestFactory()
+        self.current_conf = ConferenceFactory.create(accepting_bids=True)
+        self.current_conf.status = 'upcoming'
+        self.current_conf.save()
+        self.previous_conf = ConferenceFactory.create(accepting_bids=False)
+        self.previous_conf.status = 'completed'
+        self.previous_conf.save()
+
+        self.profile = ProfileFactory.create()
+        self.performer = PersonaFactory.create(performer_profile=self.profile,
+                                               contact=self.profile)
+        self.current_act = ActFactory.create(performer=self.performer,
+                                             submitted=True)
+        self.current_act.conference = self.current_conf
+        self.current_act.title = "Current Act"
+        self.current_act.save()
+        self.previous_act = ActFactory.create(performer=self.performer,
+                                              submitted=True)
+        self.previous_act.title = 'Previous Act'
+        self.previous_act.conference = self.previous_conf
+        self.previous_act.save()
+        self.current_class = ClassFactory.create(teacher=self.performer,
+                                                 submitted=True)
+        self.current_class.title = "Current Class"
+        self.current_class.conference = self.current_conf
+        self.current_class.save()
+        self.previous_class = ClassFactory.create(teacher=self.performer,
+                                                  submitted=True)
+        self.previous_class.conference = self.previous_conf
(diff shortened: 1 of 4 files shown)


expo/tests/gbe/test_landing_page.py
```diff
@@ -1,97 +1,215 @@
...
+            type='Volunteer')
+        current_opportunity.save()
+        previous_opportunity = GenericEventFactory.create(
+            title="Previous Volunteering",
+            conference=self.previous_conf)
+        previous_opportunity.save()
+
+        self.current_sched = Event(
+            eventitem=current_opportunity,
+            starttime=datetime(2016, 2, 5, 12, 30, 0, 0, tzinfo=pytz.utc),
```
jonkiparsky: Below, you do this without naming the argument. I like that better. Can you change this one to match?

expo/tests/gbe/test_landing_page.py
```diff
@@ -1,97 +1,215 @@
...
+        self.previous_sched = Event(
+            eventitem=previous_opportunity,
+            starttime=datetime(2015, 2, 25, 12, 30, 0, 0, pytz.utc),
+            max_volunteer=10)
+        self.previous_sched.save()
+        
+        self.current_class_sched = Event(
+            eventitem=self.current_class,
+            starttime=datetime(2016, 2, 5, 2, 0, 0, 0, pytz.utc),
+            starttime=datetime(2016, 2, 5, 2, 30, 0, 0, pytz.utc),
```
jonkiparsky: This line is duplicated. Removing it should clear up that test failure.

jonkiparsky:
Overall this looks okay. I would like to see you back out the change in your first commit, since it seems like a needless performance hit, but if you want to push back on that, I can't be arsed to argue. 

If you can clear up the test fail and the one named arg that I highlighted, I'll be good to merge.

PR #657, opened 2016-01-09
GBE-654: Comments on review volunteer list page not filtered per conference

### LINKED ISSUE #654: Volunteer review commitments on review list page should be filtered by conference
Scratch found while testing on GBE test tonight.

http://gbe2016test.gbeadmin.webfactional.com/volunteer/reviewlist

My commitments (betty) show both last years class (Bra bootcamp 1 pm on sunday) and this years class (bra buildilng at 2 pm on sunday... this year).

Not actually a regression - but a problem.

bethlakshmi:
Closes #654
- fixes a problem on 
http://localhost:8282/volunteer/reviewlist

Where a volunteer with bookings from 2015 has their booking show up.  Because the booking also don’t show date, it looks like they are current to this year, making the visual really confusing.

The fix is to screen out all commitments that are NOT a part of this conference, and also to show the date, the way we do with volunteer windows, so it is clearer what the commitment is.

### DIFF
--- a/expo/gbe/models.py
+++ b/expo/gbe/models.py
@@ -309,7 +309,7 @@ class Profile(WorkerItem):
         Otherwise, it's the same logic as schedule() below.
         '''
         events = self.schedule
-        if conference:  
+        if conference:
             conf_events = filter(
                 lambda x: x.eventitem.get_conference() == conference, events)
         else:
--- a/expo/gbe/models.py
+++ b/expo/gbe/models.py
@@ -303,17 +303,18 @@ class Profile(WorkerItem):
                  for act in acts if act.accepted == 3 and act.is_current]
         return sum([list(s) for s in shows], [])
 
-    def get_schedule(self):
+    def get_schedule(self, conference=None):
         '''
-        Gets all of a person's schedule.  Every way the actual human could
-        be committed:
-        - via profile
-        - via performer(s)
-        - via performing in acts
-        Returns schedule as a list of Scheduler.Events
-        NOTE:  Things that haven't been booked with start times won't be here.
+        Gets all schedule items for a conference, if a conference is provided
+        Otherwise, it's the same logic as schedule() below.
         '''
-        return self.schedule
+        events = self.schedule
+        if conference:  
+            conf_events = filter(
+                lambda x: x.eventitem.get_conference() == conference, events)
+        else:
+            conf_events = events
+        return conf_events
 
     @property
     def schedule(self):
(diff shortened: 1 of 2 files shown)


expo/gbe/models.py
```diff
@@ -303,17 +303,18 @@ def get_shows(self):
                  for act in acts if act.accepted == 3 and act.is_current]
         return sum([list(s) for s in shows], [])
 
-    def get_schedule(self):
+    def get_schedule(self, conference=None):
```
jonkiparsky: I can't imagine any case where we'd want to use the unparameterized schedule() function, and I can only find one place where we are using it. get_schedule and schedule should be one function, which takes a conference, and defaults to the current conference. (not to all conferences)

Since I can only find one place where the schedule() function is called (in printable_schedules) this should be a pretty trivial change.

expo/tests/gbe/test_review_volunteer_list.py
```diff
@@ -124,6 +134,90 @@ def test_review_volunteer_as_coordinator(self):
...
+            type='Volunteer')
+        current_opportunity.save()
+        booked_sched = sEvent(
+            eventitem=current_opportunity,
+            starttime=datetime(2016, 2, 6, 9, 0, 0, 0, pytz.utc),
+            max_volunteer=1)
+        booked_sched.save()
+        worker = Worker(_item=self.volunteer.profile, role='Volunteer')
+        worker.save()
+        volunteer_assignment = ResourceAllocation(
```
jonkiparsky: Should use the ResourceAllocationFactory

expo/tests/gbe/test_review_volunteer_list.py
```diff
@@ -124,6 +134,90 @@ def test_review_volunteer_as_coordinator(self):
...
+            event=booked_sched,
+            resource=worker
+        )
+        volunteer_assignment.save()
```
jonkiparsky: save() not required

expo/gbe/models.py
bethlakshmi: Actually...I was using the unparameterized schedule function in the personal schedule.  If you go to /gbe - and switch to historical view, it also shows you your old schedule.  I figured this was useful both for people booking this year's schedule and for folks like the class coordinator, @burlexpo  and I when we try to see what folks have done for us in the past.

bethlakshmi:
@jonkiparsky - found a better way in JS-- updated and retested manually for your problem.  Please rereview.

PR #658, opened 2016-01-09
GBE-656: Submenus not working in Django-CMS 

### LINKED ISSUE #656: Only two levels of menu?
Hey -- 

If you look at the menu tree on the live site, you'll see I've created a third level menu to allow people to directly access info on a show they want.  That is, the "Shows & Events" menu looks like this:

Shows & Events
----- Special Events and Parties
----- All Shows
---------- The Rhinestone Revue
---------- The Bordello
---------- The Main Event
---------- The Sunday Showcase

All of the menu items are published and have "visible in navigation" turned on.  But, if you pull down the **Shows & Events** menu, the last four (the submost level of menu) are not there.

Am I doing something wrong? Or is there some setting in the CMS which says "only display the X levels of menu" -- because I'd like to display this level.

bethlakshmi:
- submenus appear
- positioning is not horrific
- fix an existing bug in menu.tmpl - no spaces in classes mean one big
class.

To Test:
- pull the latest data from GBE Live
- make sure you have a clean browser refresh for CSS and JS
- go to Shows & Events —> All Shows

NOW - you will see that there are 4 shows, each with a working page.

The sub menu is positioned to the right of the parent menu item and the rest of the menus in the system appear in a reasonable position relative to their parents.

Prior to this fix, this did not work - submenus didn’t appear.

Closes #656

### DIFF
--- a/expo/gbe/static/styles/gbe_bootstrap.min.css
+++ b/expo/gbe/static/styles/gbe_bootstrap.min.css
@@ -3251,6 +3251,16 @@ input[type="submit"].btn-block,input[type="reset"].btn-block,input[type="button"
 	background-clip:padding-box
 }
 
+.dropdown-menu > li.dropdown > ul.dropdown-menu{
+	left: 98%;
+	top: 0px
+}
+
+.navbar > li.dropdown > .dropdown-menu{
+	top:100%;
+	left:0;
+}
+
 .dropdown-menu.pull-right{
 	right:0;
 	left:auto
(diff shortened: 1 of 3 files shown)


jonkiparsky:
Problem: when I click on a new menu, the old menu is not dismissed. This seems like a blocker to me. 
gbe-656

bethlakshmi:
yep... fair enough...

jonkiparsky:
Can't seem to find the right password to get a current live DB dump. Going to assume that you're happy with the results you're seeing and merge.

PR #700, opened 2016-01-31
GBE 680: Registration needs a checklist

bethlakshmi:
Preliminary Pull Request

@jonkiparsky - don’t approve yet, but please do start giving me comments ASAP.  

I want to test this a lot more, and hopefully write unit tests before pulling.  I also want to keep analyzing the data and make sure I covered all the conditions.

But - this is a large amount of code sprinkled in many places, so I thought an early line of conversation was warranted.

### DIFF
--- a/expo/ticketing/models.py
+++ b/expo/ticketing/models.py
@@ -1,44 +1,41 @@
-# 
-# models.py - Contains Django Database Models for Ticketing - Defines Database Schema
-# edited by mdb 8/18/2014
-#
-
 from django.db import models
 from django.contrib.auth.models import User
 from gbe.models import Conference
 from gbetext import role_options
 
-# Create your models here.
-    
+
 class BrownPaperSettings(models.Model):
     '''
-    This class is used to hold basic settings for the interface with BPT.  It should
-    contain only one row and almost never changes.
+    This class is used to hold basic settings for the interface with BPT.
+    It should contain only one row and almost never changes.
     '''
     developer_token = models.CharField(max_length=15, primary_key=True)
     client_username = models.CharField(max_length=30)
     last_poll_time = models.DateTimeField()
-    
+
     def __unicode__(self):
-        return 'Settings:  %s (%s) - %s' % (self.developer_token, self.client_username, self.last_poll_time)
+        return 'Settings:  %s (%s) - %s' % (self.developer_token,
+                                            self.client_username,
+                                            self.last_poll_time)
+
     class Meta:
-        verbose_name_plural='Brown Paper Settings'
-    
+        verbose_name_plural = 'Brown Paper Settings'
+
+
 class BrownPaperEvents(models.Model):
     '''
     This class is used to hold the BPT event list.  It defines with Brown Paper
     Ticket Events should be queried to obtain information on the Ticket Items
     above.  This information mainly remains static - it is set up info for the
     interface with BPT.
-    
+
       - include_conferece = if True this event provides tickets for all parts
             of the conference - Classes, Panels, Workshops - but not Master
             Classes
       - include_most = includes everything EXCEPT Master Classes
-      
     '''
     bpt_event_id = models.CharField(max_length=10)
-    primary = models.BooleanField(default=False)  
+    primary = models.BooleanField(default=False)
     act_submission_event = models.BooleanField(default=False,
                                                verbose_name='Act Fee')
     vendor_submission_event = models.BooleanField(default=False,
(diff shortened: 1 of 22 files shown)


bethlakshmi:
Current state update:  All code to implement is written, migrations should be clean and pep8 checking is done.
Upcoming work - I'm using 
https://docs.google.com/spreadsheets/d/12DgQCfHD7lLXd4WOjQ078gA-oBAjsvDFW2oMge9EkOg/edit#gid=1554536220

to create rules through the ticketing admin (two new areas - rule condition and ticket condition) and verifying that all of what we want will work with this weekend's version of the live code.

also want to create unit tests.

Goal = be able to get approval on pull request by Wed. on whatever state we have... possibly with tickets for testing gaps.

expo/ticketing/models.py
```diff
@@ -146,7 +157,167 @@ class Transaction(models.Model):
...
+    Ticket conditions are additive.  X purchases = X items given to the
+    buyer
+    '''
+    tickets = models.ManyToManyField(TicketItem,
+                                     blank=False)
+
+    def __unicode__(self):
+        unicode_string = ""
+        first = True
+        for ticket in self.tickets.all():
```
jonkiparsky: This would be much better as a simple join:", ".join(self.tickets.all())

expo/ticketing/models.py
```diff
@@ -146,7 +157,167 @@ class Transaction(models.Model):
...
+    checklist item under the given condition.  The usual case is that a
+    person purchased a ticket that includes an equivalent to the current
+    item and maybe more.
+    '''
+    tickets = models.ManyToManyField(TicketItem,
+                                     blank=False)
+
+    def __unicode__(self):
+        unicode_string = ""
+        first = True
```
jonkiparsky: ", ".join(self.tickets.all())

expo/ticketing/models.py
```diff
@@ -146,7 +157,167 @@ class Transaction(models.Model):
...
+    '''
+    description = models.CharField(max_length=50, unique=True)
+
+    def __unicode__(self):
+        return unicode(self.description)
+
+
+class EligibilityCondition(models.Model):
+    '''
+    This is the paremt class connecting the conditions under which a
```
jonkiparsky: s/paremt/parent

expo/ticketing/models.py
```diff
@@ -146,7 +157,167 @@ class Transaction(models.Model):
...
+    types to link the exclusion foreign key to abstract cases of this class.
+    '''
+    checklistitem = models.ForeignKey(
+        CheckListItem,
+        related_name="%(app_label)s_%(class)s")
+
+    def no_exclusion(self, held_tickets, profile, conference):
+        no_exclusion = True
+        if held_tickets:
+            for exclusion in TicketingExclusion.objects.filter(condition=self):
```
jonkiparsky: Can you use self.ticketing_ticketingexclusion_set.all() instead?  (I think that's the correct unmunged form)
Also, it would be more efficient and readable to convert this loop into an any() expression over a list comprehension.

expo/ticketing/models.py
```diff
@@ -146,7 +157,167 @@ class Transaction(models.Model):
...
+
+        return unicode_string
+
+    def is_excluded(self, held_tickets):
+        no_exclusion = True
+        for ticket in held_tickets:
+            if ticket in self.tickets.all():
+                no_exclusion = False
+
+        return no_exclusion
```
jonkiparsky: I'm not sure about the logic here. is_excluded returns True unless some ticket in held_tickets is in this exclusion's ticket set? I think you want to return `not no_exclusion`, unless I'm missing something

expo/ticketing/models.py
```diff
@@ -146,7 +157,167 @@ class Transaction(models.Model):
...
+                unicode_string = unicode(ticket)
+                first = False
+            else:
+                unicode_string += ", " + unicode(ticket)
+
+        return unicode_string
+
+    def is_excluded(self, held_tickets):
+        no_exclusion = True
+        for ticket in held_tickets:
```
jonkiparsky: again, an any() expression would help a lot here. 
I think this is what you mean:
`return any([ticket in self.tickets for ticket in held_tickets])`

expo/ticketing/models.py
bethlakshmi: I'm not following the syntax of that or where in the code to put it?  

    ", ".join(self.tickets.all())

Seems like a nice way to avoid the for loop, but I get failures when I try to paste that in as the return value.  I'm sure I'm missing something... but when I try to dig around for documentation on the join function, I don't find a whole lot out there.

expo/ticketing/models.py
bethlakshmi: Will work on going with relative name "icketing_ticketingexclusion_set" (or the working form thereof)

I'll dig into list comprehensions... but may have to pick up post-expo.

expo/ticketing/models.py
bethlakshmi: Hmm... I get your point.. the name and the logic is muddy.

yes, "is_excluded" implies that based on the arguements provided, this case should be excluded from the refernenced condition.

And... on second thought, I don't like "not_excluded" for the condition...  double negatives are just not good.  Eventually passing through here the logic works, but it'll be hell to maintain.

Will clean up name vs. logic.

bethlakshmi:
OK... looking for thought on this.  I would totally agree that:
https://github.com/bethlakshmi/GBE2/issues/711

is needed to properly pull this to mainline.

I have the following needs:
- to get this code on live by Thurs. ~1:00PM (taking the day off work tomorrow).
- to not corrupt the code base or cause problems with the team
- to not block the need to to other potential hot fixes 

I can do a pull of this branch to live... but that means that other fixes would need to be merged to this branch (ick).

Or @jonkiparsky can approve this pull request now knowing that I've assigned the #711 ticket to myself as high priority....   or I can figure out alternate strategy to running registration.

@jonkiparsky - I want to leave a final call up to you on this and not hold a gun to your head. Approve if you feel you can, please.

If not... I may be able to NOT throw it on live, but put it on test instead, pull in the live data (especially the ticketing stuff) and do the run there.... I don't love it, I think that's plan B.

jonkiparsky:
@bethlakshmi  Fine, merging.

bethlakshmi:
Thank you.  I may actually be able to fall asleep now.  I woke up this morning thinking about why unit tests aren't working...

jonkiparsky:
Heh. I've been having nightmares about code debt, myself.

PR #705, opened 2016-02-01
GBE 703: conference filter on teacher bios

### LINKED ISSUE #703: Bios are not filtered by conference
We have 3 perfectly good bio pages:

from gbe/urls.py
```
    url(r'^bios/staff/?$',
        views.bios_staff, name='bios_staff'),
    url(r'^bios/teachers/?$',
        views.bios_teachers, name='bios_teacher'),
    url(r'^/bios/volunteers/?$',
        views.bios_volunteer, name='bios_volunteer'),
```

But they don't filter by conference.

Filtering is needed on 2 directions:

1 - filter only the given role for people who are that role in THIS conference
2 - filter for the given conference in the list of things the given person is contributing - for example - Betty taught Bra Bootcamp in 2015 and Bra Workshop in 2016, only Bra Workshop should show up on the 2016 conference page.

SCOPE DOWN - this should only apply to teachers, not to staff and volunteers.  Those URLs were never ever implemented.  We should create a new ticket for next year if we want them.

bethlakshmi:
Based off other tests in the system.
closes #703

### DIFF
--- a/expo/gbe/templates/gbe/bio_list.tmpl
+++ b/expo/gbe/templates/gbe/bio_list.tmpl
@@ -11,6 +11,10 @@
      {{view_header_text | safe}}
  
    </div>
+   {% for conf in conferences %}
+     <a href = "{{ bio_url }}?conference={{ conf.conference_slug }}">
+     {{conf.conference_slug}}</a>&nbsp;|&nbsp;
+   {% endfor %}
 
    {% for teacher in bios %}
       <div class="bio_words">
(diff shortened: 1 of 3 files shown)


jonkiparsky:
There's a problem here. on my local, /bios/teachers?conference=GBE2016 returns nothing, even though I've scheduled classes. GBE2015 works, though.

BenignCremator:
On 02/01/2016 09:03 PM, Jon Kiparsky wrote:

> There's a problem here. on my local, /bios/teachers?conference=GBE2016
> returns nothing, even though I've scheduled classes. GBE2015 works, though.

I ran into that, and traced it to how the conference slug seemed to be 
returned.  I had to wander the objects dir() on the manage.py shell to 
find the right combo of things to get the right thing returned.

Hunter

jonkiparsky:
Tested with a copy of a recent backup, that seems to work.

PR #727, opened 2016-02-15
GBE-711: testing for checklist conditions

### LINKED ISSUE #711: Registration check lists needs a ton of tests
Branch GBE-680 should not be considered resolved until unit testing is finished.

I do, however, need to close #680 which is a blocker to Reg Desk opening on Friday.

I spent what time I could between Monday and now writing 1 test of test cases that only covers a fraction of the code volume.  And I resolved the comments that I could and integration tested the code.

With the need to get data setup done tomorrow afternoon and to hit run time, I will have to pull the code onto live somehow so I can get live data merged with rule setup.

The missing tests are for the following functions:

gbe.Profiles.get_roles - DONE in branch
gbe.Profiles.has_role_in_event - DONE in branch
gbe.report_views.personal_schedule = tests exist, but the changes in logic are serious enough to need many more tests, preferably some at an integration level with some serious rules/conditions and data to emulate typical expo patterns
gbe.ticketing_idd_interface.get_checklist_items_for_tickets - DONE in branch - found a bug...
gbe.ticketing_idd_interface.get_checklist_items_for_roles - DONE in branch
gbe.ticketing_idd_interface.get_checklist_items - DONE in branch
scheduler.functions.get_roles_from_scheduler-  DONE in branch
ticketing.EligibilityConditionFactory.is_excluded - DONE in branch - found a bug...

There were one or two other changes at the model level, but they are either minor enough or intuitive enough that I don't feel bad avoiding them.  These are the functions I see as pivotal to the functioning of the code with enough logical thread paths to have gotchas.

Needless to say, at the current rate (1 decently tested function per 3 hours), I will not turn this one around before the expo.

bethlakshmi:
Much more coverage.
Found 2 bugs and fixed them.
Added some reusable scheduler functions.
Closes #711

### DIFF
--- a/expo/gbe/models.py
+++ b/expo/gbe/models.py
@@ -395,7 +395,8 @@ class Profile(WorkerItem):
 
     def has_role_in_event(self, role, event):
         '''
-        Gets all of a person's roles for a conference
+        Returns True if this person has the
+        given role in the given event
         '''
         doing_it = False
         if role == "Performer":
--- a/expo/gbe/models.py
+++ b/expo/gbe/models.py
@@ -401,15 +401,16 @@ class Profile(WorkerItem):
         if role == "Performer":
             for show in self.get_shows():
                 if show.pk == event.pk:
-                    doing_it = (doing_it or True)
-        else:
-            event_workers = event.roles(role)
-            if self in event_workers:
-                doing_it = True
-            else:
-                for perf in self.get_performers():
-                    doing_it = doing_it or (perf in event_workers)
-
+                    doing_it = True
+        elif not doing_it:
+            performers = self.get_performers()
+            for person in event.roles([role]):
+                if self.pk == person._item.pk:
+                    doing_it = True
+                else:
+                    for perf in performers:
+                        if perf.pk == person._item.pk:
+                            doing_it = True
         return doing_it
 
     def __str__(self):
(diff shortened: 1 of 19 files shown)


expo/gbe/models.py
```diff
@@ -395,21 +395,23 @@ def sched_payload(self):
...
-                    doing_it = (doing_it or True)
-        else:
-            event_workers = event.roles(role)
-            if self in event_workers:
-                doing_it = True
-            else:
-                for perf in self.get_performers():
-                    doing_it = doing_it or (perf in event_workers)
-
+                    doing_it = True
```
jonkiparsky: Why not just return True here, since we have an answer?

jonkiparsky:
There seems to be a merge conflict to resolve here

expo/gbe/models.py
bethlakshmi: I see the jury is now out on the preference online (this was an interesting read:  http://stackoverflow.com/questions/36707/should-a-function-have-only-one-return-statement)

... but back when I was originally taught to code, my teachers were fairly stringent on the "one return statement per function" approach and I'm generally in the habit of it.

Having had to debug the opposite of that a number of times in different codebases and ending up feeling like a goalie with a ball that shoots between my legs when the code dumps out of the function unexpectedly, I tend to strongly prefer the 1 return statement approach in the case where the logic gets more complicated down the line.

bethlakshmi:
Merge conflict fixed.

expo/gbe/models.py
jonkiparsky: Hm. There are definitely better ways to manage complexity - this function should really be about four lines, and call two functions which would handle the two cases - but we'll worry about that when it comes to tackling the overall tech debt. #biggerfishtofry

PR #728, opened 2016-02-16
Gbe 722: Test coverage on report_views.py

jonkiparsky:
Got to 84%, saw that Betty was working on the same file, decided to pull back to avoid merge misery


expo/tests/contexts/act_techinfo_context.py
```diff
@@ -0,0 +1,46 @@
+from tests.factories.gbe_factories import (
```
bethlakshmi: Stopping here... commute almost over... hopeful to pickup here.  I want to read this carefully to get a clue on contexts.

expo/tests/contexts/act_techinfo_context.py
```diff
@@ -0,0 +1,46 @@
...
+        self.performer = performer or PersonaFactory()
+        self.act = act or ActFactory(performer=self.performer,
+                                     conference=self.conference,
+                                     accepted=3)
+        self.tech = self.act.tech
+        self.audio = self.tech.audio
+        self.lighting = self.tech.lighting
+        self.stage = self.tech.stage
+        for i in range(cue_count):
+            CueInfoFactory.create(techinfo=self.tech)
```
bethlakshmi: For realism on this, I might suggest that cue_sequence be set to "i" - in the system they are sequentially numbered starting at 1 w/in each act...

expo/tests/contexts/volunteer_context.py
```diff
@@ -0,0 +1,44 @@
...
+    EventContainerFactory,
+    ResourceAllocationFactory,
+    SchedEventFactory,
+    WorkerFactory,
+)
+
+
+class VolunteerContext():
+    def __init__(self,
+                 profile=None,
```
bethlakshmi: It doesn't look like "profile" is used anywhere in here?
Is there a way to reconcile bid.profile and profile if both arguments are passed in?  If not, maybe the contexts should be separated somehow?

expo/tests/contexts/volunteer_context.py
```diff
@@ -0,0 +1,44 @@
...
+                 bid=None,
+                 event=None,
+                 opportunity=None,
+                 conf_event=None,
```
bethlakshmi: Also "conf_event" doesn't seem to be used anywhere...

expo/tests/gbe/test_reports.py
```diff
@@ -37,15 +42,23 @@
...
+from tests.functions.gbe_functions import (
+    grant_privilege,
+    login_as,
+)
 
 
 def _create_scheduled_show_with_acts(conference=None, qty=6):
     if not conference:
-        conference = ConferenceFactory.create()
+        conference = ConferenceFactory()
```
bethlakshmi: Is there a difference between *.create() and *() ?  I see both in our code, and I'm not sure what the difference is...

bethlakshmi:
ok... finished all questions.
Mostly i want answers before pulling, but I'm a tad worried on the unused arguments in the context, since it implies that there's available input options that don't work.

expo/tests/gbe/test_reports.py
jonkiparsky: As far as I can see, there isn't. I suspect that the __init__ function is simply calling .create() and doing some jiggery-pokery to make this work, but I haven't gone spelunking in the code. 
I admit that the *() form is a little weird, in that it looks like we're instantiating a ConferenceFactory object, which we aren't, but it feels a lot cleaner to me now that I've got used to it.

expo/tests/contexts/volunteer_context.py
jonkiparsky: Good catch. I had intended to fold that in.

expo/tests/contexts/volunteer_context.py
jonkiparsky: Yeah, that can go. A thought that I never finished thinking.

expo/tests/contexts/act_techinfo_context.py
jonkiparsky: good call

expo/tests/contexts/volunteer_context.py
jonkiparsky: Fixed.

jonkiparsky:
Okay, I hope that clears up the concerns about the unused options. Good finds.

expo/tests/gbe/test_reports.py
bethlakshmi: That was my fear too... obviously the unit tests wouldn't work if that was the case...

bethlakshmi:
Thanks.  Pulling now... (commute to Akamai :) )

PR #740, opened 2016-03-14
Gbe 730: Increase coverage for gbe/views.py

jonkiparsky:
Also revised some practices in favor of nicer ones.


BenignCremator:
Tested code, and read the commits for the past 20 days.  Looks good, like the switch from objects to factories, and many of the clean-ups.  One test is failing ("FAIL: staff_area view should load"), message with it seems to indicate that this is a known error.

PR #741, opened 2016-03-20
GBE-735: Handle correctly scheduler calendar requests if no conference days exist

jonkiparsky:


BenignCremator:
Read the code, ran the tests, verified functionality.  Code looks good, mostly covers omissions in our code, and was readable and clean.  The pages I tested did not produce the 500 errors that they had previously, and I pulled the URL from error reports.

PR #745, opened 2016-03-26
GBE-744: Install ipython shell, django will select it by default

jonkiparsky:
Installing ipython on the vagrant box causes manage.py to select it by default
Locked us to version 3.0 since current versions produce much deprecation warnings. Can unlock this when we get the upgrade going.


BenignCremator:
Installed and tested the IPython shell, both free standing and with manage.py.  Works as described, merging.

PR #747, opened 2016-04-04
Gbe 746: Break up gbe/views.py into component files

jonkiparsky:
This is a pretty massive PR, but it seemed like a good time to go after it, and I think it's a huge positive step for the codebase. 
Prerequisite for moving to class-based views.


BenignCremator:
Setup vbox, ran tests, had expected errors (same as on comparison system).  Ran server and tested a few pages, seemed ok.  Reviewed code changes, seems in order.

PR #769, opened 2016-05-02
Gbe 764: Fix but NO TEST for clone_bid failure

jonkiparsky:
@bethlakshmi I'm getting a weird failure with the ClassFactory, which is preventing the test from working - the bid object doesn't show up in the db. 

We should talk about whether this should be merged to kill the bug, or if we should hold off and try to work out what the test failure is and what it means. PR issued so you can review and consider.

### DIFF
--- a/expo/gbe/models.py
+++ b/expo/gbe/models.py
@@ -1351,22 +1351,23 @@ class Class(Biddable, Event):
                                     choices=yesno_options, default="No")
 
     def clone(self):
-        new_class = Class(teacher=self.teacher,
-                          minimum_enrollment=self.minimum_enrollment,
-                          organization=self.organization,
-                          type=self.type,
-                          fee=self.fee,
-                          other_teachers=self.other_teachers,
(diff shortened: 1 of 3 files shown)


bethlakshmi:
The delete() line is clobbering the newly created class.  I can't tell why, but the conference=current_conference isn't ruling out the newly made class, so the delete() kills it.  Commenting out the delete makes the test work.

I'm in favor of committing, but it bugs me that the delete() isn't working.

Check my checkin - if you agree, let's merge.  If you disagree, let's talk more.

jonkiparsky:
@bethlakshmi  What bothers me more is that the ClassFactory is broken, so Classes are getting created without titles. I suspect this might relate to our inheritance model. Committed a slight change to make the test more strict. Feel free to merge.

bethlakshmi:
I agree with that.
This is definitely weird, because I'd have sworn that I used the Class title in other places in the code with no problem, and I definitely saw the intended value there.

PR #775, opened 2016-05-03
GBE-772: complete coverage for scheduler.calendar_view

jonkiparsky:


jonkiparsky:
calendar_view is going to need substantial revision, but for now at least all of the code is covered.

PR #784, opened 2016-05-06
GBE 778: Tests for manage_volunteer_opportunities

### LINKED ISSUE #778: Coverage for scheduler.manage_volunteer_opportunities
Time for me to embrace contexts.  My plan of action is to make a context for show volunteers to have a nice meaty piece of test data to work with.

I suspect this round will also cover:
- edit_event_display
- get_event_display_info

bethlakshmi:
Closes #778 
100% coverage on function being tested

Also fixed the cross-conference bug

And found issues with the time zones that came up as warnings during testing, so I cleaned those up.

### DIFF
--- a/expo/tests/scheduler/test_manage_volunteer_opportunities.py
+++ b/expo/tests/scheduler/test_manage_volunteer_opportunities.py
@@ -218,3 +218,64 @@ class TestEventList(TestCase):
                      response.content)
         nt.assert_in('<ul class="errorlist"><li>required</li></ul>',
                      response.content)
+
+    def test_delete_opportunity(self):
+        context = StaffAreaContext()
+        room = RoomFactory()
+        vol_opp = context.add_volunteer_opp(room=room)
+        grant_privilege(self.privileged_user, 'Scheduling Mavens')
+        login_as(self.privileged_profile, self)
+        url = reverse(self.view_name,
+                      urlconf="scheduler.urls",
+                      args=[context.sched_event.pk])
+
+        # number of volunteers is missing, it's required
+        response = self.client.post(
+            url,
+            data={'delete': 'delete',
+                  'title': 'Delete Volunteer Opportunity',
+                  'volunteer_category': 'VA0',
+                  'num_volunteers': '',
+                  'duration': '1:00:00',
+                  'day': context.conf_day.pk,
+                  'time': '10:00:00',
+                  'location': room.pk,
+                  'opp_event_id': vol_opp.eventitem.pk,
+                  'opp_sched_id': vol_opp.pk},
+            follow=True)
+        nt.assert_equal(response.status_code, 200)
+        nt.assert_not_in('Delete Volunteer Opportunity',
+                     response.content)
+        opps = EventContainer.objects.filter(parent_event=context.sched_event)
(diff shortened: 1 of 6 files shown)


jonkiparsky:
I'm getting 8 test failures, all in the new tests, all of them "LocationItem matching query does not exist"

Doing a little more investigation

jonkiparsky:
Only get this error when running the full test suite, looks like there's a dependency issue in here.

jonkiparsky:
Looks like this is another case of the issue in GBE-696

I suspect this is actually a bug down in a template somewhere, though that's just a hunch.

bethlakshmi:
Are you doing "runcoverage" or ./manage.py tests?

If you are doing "runcoverage" try dropping your DB, then syncdb and migrate.  I find that I cannnot run coverage with normal burelesque expo data in my DB.

I don't get either of these problems in my local and that does sound like the Location Item problem we've had issues with before.  I'm not sure how to fix since I can't recreate.

jonkiparsky:
I also get this when I run tests from manage.py

expo/tests/scheduler/test_manage_volunteer_opportunities.py
```diff
@@ -0,0 +1,274 @@
...
+        self.privileged_profile = ProfileFactory()
+        self.privileged_user = self.privileged_profile.user_object
+        grant_privilege(self.privileged_user, 'Volunteer Coordinator')
+
+    def test_no_login_gives_error(self):
+        url = reverse(self.view_name,
+                      urlconf="scheduler.urls",
+                      args=["1"])
+        response = self.client.get(url)
+        nt.assert_equal(response.status_code, 302)
```
jonkiparsky: Should have brought this up before merging, but we should now be using follow=True and testing to see where we get to. Just testing status code == 302 only tells us that we redirected, but doesn't actually test anything useful.

expo/tests/scheduler/test_manage_volunteer_opportunities.py
```diff
@@ -0,0 +1,274 @@
...
+
+
+class TestEventList(TestCase):
+    view_name = 'manage_opps'
+
+    def setUp(self):
+        self.client = Client()
+        self.user = ProfileFactory.create().user_object
+        self.privileged_profile = ProfileFactory()
+        self.privileged_user = self.privileged_profile.user_object
```
jonkiparsky: Again, I failed in my review, but this object is not required since grant_privilege takes either a user or a profile.

expo/tests/scheduler/test_manage_volunteer_opportunities.py
```diff
@@ -0,0 +1,274 @@
...
+        context = StaffAreaContext()
+        room = RoomFactory()
+        grant_privilege(self.privileged_user, 'Scheduling Mavens')
+        login_as(self.privileged_profile, self)
+        url = reverse(self.view_name,
+                      urlconf="scheduler.urls",
+                      args=[context.sched_event.pk])
+        response = self.client.post(
+            url,
+            data={'create': 'create',
```
jonkiparsky: Another fail on my part. This duplicated code should have been a function.

PR #787, opened 2016-05-14
GBE-785: 100% coverage for add_event

### LINKED ISSUE #785: test coverage for sheduler.add_event
Tests for add event.

bethlakshmi:
closes #785 
- 100% coverage of add_event
- fixed the testability problem in the EventScheduler Form
- found a bug in add_event (no staff lead allocation)
- created a new context - PanelContext
- put some more reality into Class and GenericEvent Contexts to better simulate the normal case for these data sets.
- pep8 here and there

### DIFF
--- a/expo/scheduler/views.py
+++ b/expo/scheduler/views.py
@@ -773,7 +773,8 @@ def add_event(request, eventitem_id, event_type='Class'):
     '''
     profile = validate_perms(request, ('Scheduling Mavens',))
 
-    item = get_object_or_404(EventItem, eventitem_id=eventitem_id)
+    eventitem = get_object_or_404(EventItem, eventitem_id=eventitem_id)
+    item = eventitem.child()
     template = 'scheduler/event_schedule.tmpl'
     eventitem_view = get_event_display_info(eventitem_id)
 
--- a/expo/scheduler/views.py
+++ b/expo/scheduler/views.py
@@ -494,6 +494,7 @@ def allocate_workers(request, opp_id):
                                         urlconf='scheduler.urls',
                                         args=[opp.event_type_name, opp_id]))
 
+
 @login_required
 def manage_volunteer_opportunities(request, event_id):
     '''
@@ -514,9 +515,9 @@ def manage_volunteer_opportunities(request, event_id):
     if request.method != 'POST':
         # TO DO: review this
         return HttpResponseRedirect(reverse('edit_event',
-                                        urlconf='scheduler.urls',
-                                        args=[event.event_type_name,
-                                                event_id]))
+                                    urlconf='scheduler.urls',
+                                    args=[event.event_type_name,
+                                          event_id]))
     if 'create' in request.POST.keys() or 'duplicate' in request.POST.keys():
         if 'create' in request.POST.keys():
             form = VolunteerOpportunityForm(
@@ -535,7 +536,8 @@ def manage_volunteer_opportunities(request, event_id):
             data = form.cleaned_data
             day = data.get('day').day
             time_parts = map(int, data.get('time').split(":"))
-            start_time = datetime.combine(day, dttime(*time_parts, tzinfo=pytz.utc))
+            start_time = datetime.combine(day, dttime(*time_parts,
+                                                      tzinfo=pytz.utc))
             opp_event = Event(eventitem=opp.eventitem_ptr,
                               max_volunteer=data.get('num_volunteers', 1),
                               starttime=start_time,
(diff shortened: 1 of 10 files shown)


expo/tests/scheduler/test_add_event.py
```diff
@@ -1,100 +1,331 @@
...
+        context.add_panelist()
+        overcommitter1 = PersonaFactory()
+        overcommitter2 = PersonaFactory()
+        login_as(self.privileged_profile, self)
+        url = reverse(self.view_name,
+                      urlconf="scheduler.urls",
+                      args=["Class", context.bid.eventitem_id])
+        response = self.client.post(
+            url,
+            data={'event-day': context.days[0].pk,
```
jonkiparsky: Please extract this repeated code to a function (many examples in tests/gbe)

expo/tests/scheduler/test_add_event.py
```diff
@@ -1,100 +1,331 @@
...
+        self.privileged_user = self.privileged_profile.user_object
+        grant_privilege(self.privileged_user, 'Scheduling Mavens')
+        self.eventitem = GenericEventFactory()
+
+    def test_no_login_gives_error(self):
+        url = reverse(self.view_name,
+                      urlconf="scheduler.urls",
+                      args=["GenericEvent", self.eventitem.eventitem_id])
+        response = self.client.get(url)
+        nt.assert_equal(response.status_code, 302)
```
jonkiparsky: Where does the redirect take us? Please use a follow=True and test this to its final destination

expo/tests/scheduler/test_add_event.py
```diff
@@ -1,100 +1,331 @@
...
+        nt.assert_equal(response.status_code, 404)
+
+    def test_good_user_get_success(self):
+        login_as(self.privileged_profile, self)
+        url = reverse(self.view_name,
+                      urlconf="scheduler.urls",
+                      args=["GenericEvent", self.eventitem.eventitem_id])
+        response = self.client.get(url)
+        nt.assert_equal(response.status_code, 200)
+        nt.assert_in(self.eventitem.title,
```
jonkiparsky: This is good.

expo/tests/scheduler/test_add_event.py
```diff
@@ -1,100 +1,331 @@
...
+                  },
+            follow=True)
+
+        assert_redirects(response, reverse('event_schedule',
+                                           urlconf='scheduler.urls',
+                                           args=["Class"]))
+
+        nt.assert_not_in('<ul class="errorlist">', response.content)
+        nt.assert_in('Events Information', response.content)
+        sessions = Event.objects.filter(eventitem=context.bid, max_volunteer=3)
```
jonkiparsky: I prefer to make this a comparison to the initial state of the object. That way, if someone decides to change the behavior of the context you're using here, it's not going to break this test. 
Also, `sessions = context.bid.scheduler_events.filter(max_volunteer=3)` is a little nicer way to say the same thing, and obviates the need to import Event
It would be nice also to see this generalized into a function, but that's not a blocker.

expo/tests/scheduler/test_add_event.py
```diff
@@ -1,100 +1,331 @@
...
+        nt.assert_equal(len(sessions), 1)
+        session = sessions.first()
+        nt.assert_equal(session.starttime,
+                        datetime.combine(context.days[0].day,
+                                         time(12, 0, 0, tzinfo=pytz.utc)))
+        nt.assert_in('New Title', response.content)
+        nt.assert_in(str(context.room), response.content)
+        nt.assert_in('3', response.content)
```
jonkiparsky: For how many of our pages would this assertion fail? Surely the character '3' appears on a large number of our pages. Can we either make this more specific or remove it?

expo/tests/scheduler/test_add_event.py
```diff
@@ -1,100 +1,331 @@
...
+                  'event-description': 'New Description',
+                  })
+
+        nt.assert_equal(response.status_code, 200)
+        nt.assert_in('<input id="id_event-title" name="event-title" ' +
+                     'type="text" value="New Title" />',
+                     response.content)
+        nt.assert_in("New Description",
+                     response.content)
+        nt.assert_in('<input id="id_event-max_volunteer" min="0" ' +
```
jonkiparsky: This looks more like it. This makes me want to write some testing utils functions to handle working with the html more cleanly, but I'm not going to push for that here.

expo/scheduler/views.py
```diff
@@ -810,6 +808,11 @@ def add_event(request, eventitem_id, event_type='Class'):
                 for panelist in data['panelists']:
                     s_event.allocate_worker(panelist.workeritem, 'Panelist')
 
+            if data['staff_lead']:
```
jonkiparsky: Clearly, this is crying out to be a function.

bethlakshmi: Yes, but I'd rather do that refactor AFTER we have tests on edit_event.

bethlakshmi:
@jonkiparsky - OK, I think I got all the comments that are blockers.

PR #789, opened 2016-05-15
Gbe 742: Replaces nose with pytest as test runner

jonkiparsky:
This simply replaces the mechanism without adding or removing any functionality. 

The main difference is that instead of calling manage.py to run the tests, we run 

```py.test /PATH/TO/tests/```

Obviously the path specification works as usual, ie if we're in /vagrant/expo, it's just `py.test tests`

Of course we can specify more precisely, eg
```py.test /PATH/TO/tests/scheduler```
```py.test /PATH/TO/tests/gbe/test_bid_acts.py```

The other thing to note is that I've set pytest up to reuse the db from the last run. Since it handles rollbacks, this will only matter if you have made a model change since the last time you ran the tests. In that case you'd run

```py.test --create-db /PATH/TO/tests```

the `runcoverage` alias has been replaced to run under pytest - should work correctly. @bethlakshmi  please test this with a dump from live or test loaded into gbe_dev, if this doesn't work correctly I want to know about it!

We can talk in more depth about the changes to the tests themselves, but the main ones are:
- if a test touches the database it must either be marked for django-db (as in tests/scheduler/test_view_list.py) or it must be in a test case class that inherits from django.test.TestCase. 
- we can write nose asserts if we want, but standard python assert should work perfectly, and we should prefer them unless there's some reason to use the nose assert. 

Finally, note that this change touches requirements.txt, so it will be necessary to rebuild your vagrant environment to test this.

### CODE REVIEW

**Comment by bethlakshmi** [comment]
OK... something's not right.  I did the following (just to start clean, after several errors):
```
vagrant reload --provision
vagrant up
vagrant ssh
cd /vagrant/expo
./manage.py syncdb (set up an admin...)
./manage.py migrate
```

then...
`py.test /vagrant/expo/tests`

is giving me tons of failures.

the final outcome was:

`===================== 311 failed, 56 passed, 4 error in 240.44 seconds =====================
`

Then, just to see what happens, I ran:

`usage: py.test [options] [file_or_dir] [file_or_dir] [...]`

and got...

`py.test: error: unrecognized arguments: --create_db`

runcoverage is having the same problems as regular testing ... so I didn't try the live DB test yet - so many things are failing, it would be hard to see if the state changed...

**Comment by jonkiparsky** [comment]
I see the problem. To deal with the errors, you'll have to add the following to your local_settings.py databases settings:
```
DATABASES = {
  ...
        'OPTIONS': {  
            "init_command": "SET foreign_key_checks = 0;",
}
```

Also, it's create-db, not create_db.

**Comment by bethlakshmi** [comment]
Still not working.

I get lots of the error:

`E       OperationalError: (1049, "Unknown database 'test_gbe_dev'")
`

Here's my local settings:
[local_settings.txt](https://github.com/bethlakshmi/GBE2/files/265238/local_settings.txt)

And the errors I'm getting:

[pytest_out.txt](https://github.com/bethlakshmi/GBE2/files/265239/pytest_out.txt)

Do you do something to define test_gbe_dev?
I hadn't had to with nose...

**Comment by bethlakshmi** [comment]
This time for real...

I did...
`vagrant reload --provision
vagrant up
vagrant ssh
cd /vagrant/expo
./manage.py syncdb (set up an admin...)
./manage.py migrate
`

And ran...
`py.test --create-db /vagrant/expo/tests`

And got:

1 error, 2 failures - 
[pytest_out.txt](https://github.com/bethlakshmi/GBE2/files/265264/pytest_out.txt)


My local settings are:

`DEBUG = True
TEMPLATE_DEBUG = False

LOGIN_REDIRECT_URL = '/'
#ALLOWED_HOSTS = ['localhost', '127.0.0.1']

STATIC_ROOT = '/vagrant/expo/gbe/static'
STATIC_URL = "/static/"


# mc keys omitted

USER_CONTACT_RECIPIENT_ADDRESSES = 'betty@burlesque-expo.com'

SEND_NO_MAIL = True

DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.mysql',
        'NAME': 'gbe_dev',
        'USER': 'django_user',
        'PASSWORD': 'secret',
        'HOST': '',
        'PORT': '',
        'OPTIONS': {  
            "init_command": "SET foreign_key_checks = 0;",
        }
    },

}
`

**Comment by jonkiparsky** [comment]
I see you're doing a reload instead of a destroy. First thing I would try in this case would be a vagrant destroy/vagrant up. If that doesn't do it, we can dive deeper.

**Comment by bethlakshmi** [comment]
Wohoo!  - OK - needed to do:
```

vagrant destroy
vagrant up
vagrant reload --provision
vagrant up
vagrant ssh
cd /vagrant/expo
./manage.py syncdb (set up an admin...)
./manage.py migrate
```
I'm not sure both up commands were needed, but it worked, I'm error free.

I'll read up a bit on the checkin and test run coverage with a seeded db before approving.

File: expo/tests/scheduler/test_view_list.py [hunk 1]
```diff
@@ -84,24 +87,27 @@ def test_view_list_default_view_current_conf_exists():
...
+                        urlconf="scheduler.urls",
+                        args=[param.lower()])
+
+    url_upper = reverse("event_list",
+                        urlconf="scheduler.urls",
+                        args=[param.upper()])
+    user = ProfileFactory.create().user_object
+    client.login(username=user.username,
+                 email=user.email,
+                 password=password)
```
bethlakshmi: Actually - why login at all here?  This is a public function with no login required.

bethlakshmi:
Uploaded my problem child DB image.  "runcoverage" works!

Two questions 
1 - see comment

2 - when is it needed to use "pytest.mark.django_db" and when can it be skipped?  I see you putting it in in places where the code was probably failing... but I can't when it's the right choice.  The test says "This is used to mark a test function as requiring the database" - but ... don't almost all our test cases require the DB?

Other than those two questions - I'm good to approve.

expo/tests/scheduler/test_view_list.py
jonkiparsky: Good call - force of habit. I'll remove that.

jonkiparsky:
pytest does a little magic to override the need for mark.django_db if a test case is a subclass of django.test.TestCase. This is why the import of unittest.TestCase was changed to django.test universally.

jonkiparsky:
@bethlakshmi If you want to get this merged, feel free to remove that gratuitous login line and merge (assuming tests still pass)

PR #796, opened 2016-05-18
added test case for when no conference slug is passed in

gawaineo:


bethlakshmi:
Looks good!
And now it's 100%!

PR #797, opened 2016-05-20
GBE 783: allocate_worker coverage

### LINKED ISSUE #783: test coverage for scheduler.allocate_worker
Should also hit edit_event_display

bethlakshmi:
Closes #783 

I found a lot of bugs in the error handling area, and cleaned them up and wrote test cases.  Because the volunteer allocation form doesn’t really let a user put in bad data (unless you hack it with Inspect Element) - these weren’t evident.  But they CAN be exercised via integration test… so I did.


jonkiparsky:
PEP-8:

```expo/tests/scheduler/test_allocate_workers.py:10:44: E261 at least two spaces before inline comment
expo/tests/scheduler/test_allocate_workers.py:16:1: E302 expected 2 blank lines, found 1
expo/tests/scheduler/test_allocate_workers.py:57:5: E303 too many blank lines (2)
expo/tests/scheduler/test_allocate_workers.py:71:5: E303 too many blank lines (2)
expo/tests/scheduler/test_allocate_workers.py:79:32: E261 at least two spaces before inline comment
```

.cache/v/cache/lastfailed
```diff
@@ -0,0 +1 @@
+{}
```
jonkiparsky: remove

expo/scheduler/templates/scheduler/allocate_volunteers.tmpl
```diff
@@ -35,8 +32,8 @@ included in event edit flow. Show WorkerAllocationForms
 	  {{field}}
 	{% endfor %}
 	  <td>  <input type = "submit" name="edit" value="Edit/Create Allocation"></td>
-	  {% if form.alloc_id.value > 0  %}
-	  <td>  <input type = "submit" name="delete" value="Delete Allocation"></td>
+	  {% if form.alloc_id.value > 0 and form.alloc_id.value != '-1' %}
```
jonkiparsky: Shouldn't this be addressed upstream? If we're sometimes passing an int and sometimes we're passing the string "-1", then that's pretty bad. If we're not going to fix this in this PR, then this should be filed as an issue.

bethlakshmi:
@jonkiparsky - all comments addressed.

PR #802, opened 2016-05-24
GBE 801: Add test case for review_vendors_list_view

gawaineo:
test case asserts that all vendors created with the VendorFactory appear on the response content and also that we received a 200 status.


jonkiparsky:
Looks great!

PR #807, opened 2016-05-28
GBE 805: full coverage for edit act techinfo

### LINKED ISSUE #805: Full coverage for gbe_views_edit_act_techinfo_view.py and below
There's some uncovered code in gbe.models used here, so I'm trying to get both areas covered.

bethlakshmi:
Closes #805 
minor bugs
lots of tests
got as much of the models and forms as I could.


expo/gbe/views/edit_act_techinfo_view.py
```diff
@@ -77,20 +78,30 @@ def EditActTechInfoView(request, act_id):
             except:
                 pass
             re_set.insert(0, existing_rehearsal)
+            re_set = sorted(re_set,
```
jonkiparsky: If you're sorting the set, then the insert should be replaced by an append. 
Also, the sort should happen outside the loop.

jonkiparsky: If you're sorting the set, then the insert should be replaced by an append. 
Also, the sort should happen outside the loop.

expo/gbe/views/edit_act_techinfo_view.py
```diff
@@ -77,20 +78,30 @@ def EditActTechInfoView(request, act_id):
...
     if len(rehearsal_sets) > 0:
-        rehearsal_forms = [RehearsalSelectionForm(
-            initial={'show': show,
-                     'rehearsal_choices':
-                     [(r.id, "%s: %s" % (
-                         r.as_subtype.title,
-                         r.starttime.strftime("%I:%M:%p"))) for r in r_set]})
-                           for (show, r_set) in rehearsal_sets.items()
-                       ]
+        rehearsal_forms = []
```
jonkiparsky: This should be extracted to a function

expo/tests/contexts/act_techinfo_context.py
```diff
@@ -50,17 +50,18 @@ def __init__(self,
...
-    rehearsal_event = SchedEventFactory(eventitem=rehearsal.eventitem_ptr,
-                                        max_volunteer=10)
-    event_container = EventContainerFactory(
-        child_event=rehearsal_event,
-        parent_event=s_event)
-    ResourceAllocationFactory(resource=ActResourceFactory(
-        _item=act.actitem_ptr),
-                              event=s_event)
-    return rehearsal_event
+    def _schedule_rehearsal(self, s_event, act=None):
```
jonkiparsky: I don't see where you're using the self. Why is this being moved into the class?

expo/gbe/views/edit_act_techinfo_view.py
bethlakshmi: I fixed the insert --> append, but I disagree on the loop.

The for loop is iterating through shows, there is 1 rehearsal set (re_set) per show, and each set should be sorted separately.  It only needs sorting if the act has an existing rehearsal choice (that's when the existing sorted order gets disrupted). 

Reviewing it, though, I found a case where sort should execute less frequently, and there's less screwing with the list, for the same outcome.  Now it will only re-sort the list when the rehearsal that the person is signed up for had to be added to the list - that happens when the rehearsal was fully booked.

PR #809, opened 2016-05-28
GBE-808: Added missing migration

jonkiparsky:
Note that this will conflict with current master, so need to run it as --fake if you're testing against a recent copy of the prod db


bethlakshmi:
So, I just ran:
`./manage.py migrate --fake`
`./manage.py migrate`

and it seems fine.  But I don't understand why we need it?

The migration references costume, cue sheet and lighting info - all of which are from the time before we had migrations.  What makes these special that we need to add migrations for them?

jonkiparsky:
Let's talk about this in realspace. This might not be necessary - but I've
been running into some weird migrations issues.

PR #811, opened 2016-05-29
Gbe 810 - Completed test coverage for create_volunteer_view.py

gawaineo:


expo/tests/gbe/test_create_volunteer.py
```diff
@@ -78,3 +80,18 @@ def test_create_volunteer_no_post(self):
...
         response = self.client.post(url)
         nt.assert_equal(response.status_code, 200)
+
+    def test_create_volunteer_post_with_submit_is_true(self):
+        url = reverse(self.view_name,
+                      urlconf='gbe.urls')
+        login_as(ProfileFactory(), self)
+        data = self.get_volunteer_form(submit=True)
+        response = self.client.post(url, data=data)
+        nt.assert_equal(response.status_code, 302)
```
jonkiparsky: It would be better to add a "follow=True" clause to the client.post() call and then test the content of the response. Also, since it's a post we should expect it to attempt to affect the contents of the db, and we should test those changes. In this case, the test might be to get a count of Volunteers in the database (`Volunteer.objects.count()`) prior to the post, and compare that value to the same call after the post. Expect the value to increase by 1.

expo/tests/gbe/test_create_volunteer.py
```diff
@@ -78,3 +80,18 @@ def test_create_volunteer_no_post(self):
...
+
+    def test_create_volunteer_with_get_request(self):
+        url = reverse(self.view_name,
+                      urlconf='gbe.urls')
+        login_as(ProfileFactory(), self)
+        response = self.client.get(url)
+        nt.assert_equal(response.status_code, 200)
```
jonkiparsky: Can we test response.content?

expo/tests/gbe/test_create_volunteer.py
```diff
@@ -25,15 +25,17 @@ def setUp(self):
         self.performer = PersonaFactory()
         Conference.objects.all().delete()
         self.conference = ConferenceFactory(accepting_bids=True)
-        days = ConferenceDayFactory.create_batch(2, conference=self.conference)
+        days = ConferenceDayFactory.create_batch(3, conference=self.conference)
```
jonkiparsky: I'm curious about the motivation for this change.

gawaineo: Below in the get_volunteer_form() function I do some slicing for the available and unavailable windows. The intention was to use the third ConferenceDay as an unavailable window for when the form is being submitted.

```python
def get_volunteer_form(self, submit=False, invalid=False):
        form = {'profile': 1,
                'number_shifts': 2,
                'interests': ['VA0'],
                'available_windows': self.conference.windows().values_list(
                    'pk', flat=True)[0:2],
                'unavailable_windows': self.conference.windows().values_list(
                    'pk', flat=True)[2]
                }
        if submit:
            form['submit'] = True
        if invalid:
            del(form['number_shifts'])
        return form
```

jonkiparsky: Okay, cool. Thanks.

expo/tests/gbe/test_create_volunteer.py
```diff
@@ -85,10 +85,12 @@ def test_create_volunteer_post_with_submit_is_true(self):
         url = reverse(self.view_name,
                       urlconf='gbe.urls')
         login_as(ProfileFactory(), self)
+        num_volunteers_before = Volunteer.objects.count()
         data = self.get_volunteer_form(submit=True)
         response = self.client.post(url, data=data, follow=True)
         nt.assert_equal(response.status_code, 200)
         nt.assert_in('Profile View', response.content)
+        nt.assert_equal(num_volunteers_before + 1, Volunteer.objects.count())
```
jonkiparsky: Nice.

expo/tests/gbe/test_create_volunteer.py
```diff
@@ -96,3 +96,4 @@ def test_create_volunteer_with_get_request(self):
         login_as(ProfileFactory(), self)
         response = self.client.get(url)
         nt.assert_equal(response.status_code, 200)
+        nt.assert_in('Volunteer', response.content)
```
jonkiparsky: Okay.

expo/tests/gbe/test_create_volunteer.py
```diff
@@ -86,8 +86,9 @@ def test_create_volunteer_post_with_submit_is_true(self):
                       urlconf='gbe.urls')
         login_as(ProfileFactory(), self)
         data = self.get_volunteer_form(submit=True)
-        response = self.client.post(url, data=data)
-        nt.assert_equal(response.status_code, 302)
+        response = self.client.post(url, data=data, follow=True)
+        nt.assert_equal(response.status_code, 200)
+        nt.assert_in('Profile View', response.content)
```
jonkiparsky: Okay, this should work.

jonkiparsky:
Looks good, should be able to merge as soon as I can rebuild my context.

PR #816, opened 2016-06-05
Revert "No reason for this?"

jonkiparsky:
This reverts commit 50ed0aa8d4aa6d5851910ef9059485a618837f54 which deleted the file scheduler/table.py

### DIFF
--- a/expo/scheduler/table.py
+++ b/expo/scheduler/table.py
@@ -1,0 +1,201 @@
+class table:
+    '''
+    A table datatype class for Python.
+
+    It accepts rows and columns, and returns the data object stored in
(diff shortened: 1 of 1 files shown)


bethlakshmi:
Why?  Where is it used?  I couldn't find any way to test it, because I couldn't find a place that used or included it.

jonkiparsky:
do a recursive delete of *.pyc from your local working copy, then run tests. 

also, in a python shell try to import "table" and see what you get.

jonkiparsky:
for where it's used, see expo/scheduler/functions.py, line 1:

```
from table import table
```

PR #824, opened 2016-06-19
GBE 817 - fix for errors in Act Tech Info submit cause messed up information display on top of page.

### LINKED ISSUE #817: When you input an act tech info and get an error it shows you too much data in your summary
I just did this:
- went to an act tech info for an act in the active conference as a performer in that act
- edited it in a way that causes a form error
- my summary when from a few brief lines to a hemmorage of information because it shows every performer in the system in the view pain.

This is an information leak and a bug.  I will fix it.

bethlakshmi:
Makes the failing test pass.
Refactored code to use the same act info setup code regardless of post
or get, and used the working get process instead of the post process.

A way to see the problem in integration is to follow the steps of #817 before and after the fix.

closes #817

### DIFF
--- a/expo/gbe/views/edit_act_techinfo_view.py
+++ b/expo/gbe/views/edit_act_techinfo_view.py
@@ -34,67 +34,6 @@ from gbe.forms import (
 from scheduler.models import Event as sEvent
 
 
-def set_rehearsal_forms(shows, act):
-    '''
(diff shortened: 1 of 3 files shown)


jonkiparsky:
@bethlakshmi Changes look good I'm getting a merge conflict here. Please resolve that and push. Thanks!

PR #825, opened 2016-06-22
GBE-806: Resolve doubly-inherited fields

jonkiparsky:
For review, need to discuss before merging. 

This is the revision to gbe.Biddable and gbe.Event that we discussed: the three conflicting fields got renamed to distinguish them. 

Migrations may be a bit wonky - you might have to --fake 0007, for example - but I'm not too worried about them since they'll be deleted as soon as we get up to the >1.7

Please review and look for problems. Extra points if you give me failing cases where Classes are getting out of sync on the duplicated fields!

DO NOT MERGE until we've hashed this out out!

### DIFF
--- a/expo/expo/settings.py
+++ b/expo/expo/settings.py
@@ -130,9 +130,9 @@ INSTALLED_APPS = (
     #    'djangocms_teaser',
     #    'djangocms_video',
     'reversion',  # for versioning in cms -- use easy install
-    'gbe',
-    'ticketing',
     'scheduler',
+    'ticketing',
+    'gbe',
     'django_nose',
     'hijack',
     'compat',
--- a/expo/gbe/models.py
+++ b/expo/gbe/models.py
@@ -659,7 +659,7 @@ class AudioInfo(models.Model):
 
     def __unicode__(self):
         try:
-            return "AudioInfo: " + self.techinfo.act.title
+            return "AudioInfo: " + self.techinfo.act.b_title
         except:
             return "AudioInfo: (deleted act)"
 
@@ -699,7 +699,7 @@ class LightingInfo (models.Model):
 
     def __unicode__(self):
         try:
-            return "LightingInfo: " + self.techinfo.act.title
+            return "LightingInfo: " + self.techinfo.act.b_title
         except:
             return "LightingInfo: (deleted act)"
 
@@ -758,7 +758,7 @@ class StageInfo(models.Model):
 
     def __unicode__(self):
         try:
-            return "StageInfo: " + self.techinfo.act.title
+            return "StageInfo: " + self.techinfo.act.b_title
         except:
             return "StageInfo: (deleted act)"
 
@@ -807,7 +807,7 @@ class TechInfo(models.Model):
 
     def __unicode__(self):
         try:
-            return "Techinfo: " + self.act.title
+            return "Techinfo: " + self.act.b_title
         except:
             return "Techinfo: (deleted act)"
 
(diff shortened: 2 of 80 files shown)


bethlakshmi:
I see this... I'll look tomorrow! ;)  Can't wait to see it.

bethlakshmi:
OK... I'm definitely getting lost trying to run the tests.  I decided not to fool around with existing DBs, I just started clean... I wiped my DB, made sure I had created the gbe_dev database.

I did:

`./manage.py syncdb `

Worked fine, no problems.

`./manage.py migrate`

Looked good, I see the new migrations (0007, 0008, 0009, 0010), and " - Migration 'gbe:0009_migrate_event_and_biddable_fields' is marked for no-dry-run."

Then I do:

`runcoverage`

And the output is attached as a file
[db-isolation-test-fails.txt](https://github.com/bethlakshmi/GBE2/files/342743/db-isolation-test-fails.txt)

jonkiparsky:
runcoverage is set up to reuse the existing db.  Try

`py.test --create-db tests`

expo/expo/settings.py
```diff
@@ -130,9 +130,9 @@
     #    'djangocms_teaser',
     #    'djangocms_video',
     'reversion',  # for versioning in cms -- use easy install
-    'gbe',
-    'ticketing',
     'scheduler',
+    'ticketing',
+    'gbe',
```
bethlakshmi: Out of curiosity - why does this matter?

expo/gbe/models.py
```diff
@@ -36,8 +36,8 @@
 
 phone_regex = '(\d{3}[-\.]?\d{3}[-\.]?\d{4})'
 
-visible_bid_query = (Q(biddable_ptr__conference__status='upcoming') |
-                     Q(biddable_ptr__conference__status='ongoing'))
+visible_bid_query = (Q(biddable_ptr__b_conference__status='upcoming') |
+                     Q(biddable_ptr__b_conference__status='ongoing'))
```
bethlakshmi: given that this works now with bids and not events, would there be a way to put this within the parent Bid class?

bethlakshmi:
running the unit tests isn't enough, we need to run and look at each page where we've edited the model.
So far:
- I fixed the points in gbe/admin.py that prevented the server from running.
- I looked at bid_class and the labels are messed up - they now show b_title, and b_description

We need to review all of the following pages:
- bid, edit, review class
- bid, edit, review act
- bid, edit, review vendor
- bid, edit, review, assign volunteer
- bid, edit, review costume
- create a new event - generic, show, class
- schedule event
- edit a scheduled event
- volunteer allocation
- the reports.  Probably all of them.

to make sure that b_ or e_ isn't showing up for users.

bethlakshmi:
I'll keep digging into the structures of the code (I'm at maybe 9 out of 80 files), but the previous comment is a blocker to pulling the change.

bethlakshmi:
amend the previous comment - we can't just view the pages, we've got to run integration tests.

I did the following:
1 - go to http://localhost:8282/class/create
2 - enter a "b_title" and nothing else
3 - save as draft

Get error:

IntegrityError at /class/create

(1048, "Column 'b_conference_id' cannot be null")

-----

I repeated the same test for submitting for approval (also gave a preferred teaching time) and that failed as well.

We need to run manual integration tests here, and I'd be thrilled if we found a way to cover these really basic tests in the unit tests (are these tests that we marked to skip?) - I'm actually surprised that these weren't already done as part of code coverage...

I'm going to pause here, and let @jonkiparsky update the code before more review.

expo/expo/settings.py
jonkiparsky: Because under 1.7, ticketing has to load before gbe.

PR #832, opened 2016-07-10
GBE-831: Lock down cmsplugin-filer and django-nose to versions that

jonkiparsky:
support our configuration. 

To test, pull a new clone of the repo. Stand up the vagrant box and attempt to run tests or get into the django shell - should fail. Then check out this branch and do a 
`vagrant reload --provision
`
or 
`vagrant destroy
vagrant up
`

Then ssh in and run tests. Should succeed.


bethlakshmi:
works for me.

PR #835, opened 2016-07-13
GBE 823:  User alert messages when submission of a state change is accomplished.

### LINKED ISSUE #823: Create Event Based messaging system for bids
**Since this is going to apply to class and act bids, it needs to be up and running by 7/31/16** When a user successfully submits or edits a volunteer bid, that user should get a pop-up which corresponds to the action they have taken.

Tech details below.

bethlakshmi:
Closes #823 
- moved the message alert popup to the base.tmpl and made it bootstrap alerts
- had to refactor the css for our other alerts to avoid conflict of styles
- for every time the regular end user changes a state, there is now an alert message.
- alert messages show up in admin in the db and have a migration 
- view and code can’t be edited by an admin to protect from breaking the code.
- alerts have a default in case the user has deleted them.
- tests for everything
- added some additional tests to provide more coverage on tested views.
- found a bug in a previous fix to edit act test info - fixed the fix and made a test for it


expo/gbe/migrations/0008_auto__add_usermessage__add_unique_usermessage_view_code.py
```diff
@@ -0,0 +1,387 @@
+# -*- coding: utf-8 -*-
```
jonkiparsky: Migration numbering is off.

expo/gbe/migrations/0007_auto__add_costume__add_field_show_cue_sheet__add_field_lightinginfo_sp.py
```diff
@@ -1,396 +0,0 @@
-# -*- coding: utf-8 -*-
```
jonkiparsky: This migration was already deleted. Please make sure your repo is up to date!

jonkiparsky:
Clearly we're going to have to do that refactoring of the duplicated bid logic sooner rather than later. Not a blocker on this commit, but this PR is almost impossible to review, because all of the significant changes are buried in repetitive boilerplate.

expo/tests/gbe/test_bid_costume.py
```diff
@@ -61,37 +88,46 @@ def test_costume_bid_post_form_not_valid(self):
...
-                      urlconf="gbe.urls")
-        login_as(PersonaFactory().contact, self)
-        data = self.get_costume_form(submit=True)
-        response = self.client.post(url, data=data, follow=True)
-        nt.assert_equal(response.status_code, 200)
-        nt.assert_true(('http://testserver/gbe', 302)
-                       in response.redirect_chain)
-        nt.assert_true("Your Account" in response.content)
+        response, data = self.post_costume_submission()
+        self.assertEqual(response.status_code, 200)
```
jonkiparsky: Why, why, why are you wasting time with these changes? nt asserts are fine, changing them just makes it harder to find the substantive changes - and if you're going to change them, we should be moving to naked python asserts, not someone else's assert framework!

http://docs.pytest.org/en/latest/assert.html

expo/gbe/migrations/0008_auto__add_usermessage__add_unique_usermessage_view_code.py
bethlakshmi: Yes, and I'm honestly afraid to change it - there's something really weird happening with the gbe_costumes - I had to make a migration (the 0007 migration) to get this migration to work... but the I had to fake it or I couldn't run 0008 and then I had to delete it, or the unit tests wouldn't run.

expo/tests/gbe/test_bid_costume.py
bethlakshmi: When I read:

https://github.com/bethlakshmi/GBE2/pull/789

My takeaway was that we should use standard python asserts.  I assumed that asserts coming from the django.test.TestCase would be standard enough and that it would be worthwhile to clear out the dependancy on nose.

Fine.  Will stop it in the future - do you want me to pull that part of these changes?

expo/gbe/migrations/0008_auto__add_usermessage__add_unique_usermessage_view_code.py
jonkiparsky: Okay, I'm going to call this a "whatever". Either we're going to get up to supported django and delete all of these migrations, or we're not and it won't matter. Not a blocker.

expo/tests/gbe/test_bid_costume.py
jonkiparsky: Well, I did say exactly the opposite of that:

> we can write nose asserts if we want, but standard python assert should work perfectly, and we should prefer them unless there's some reason to use the nose assert.

pytest understands an assertion like `assert paul in beatles` and will raise a reasonable error if it turns out that paul is not a member of that set. 

No, don't bother undoing, but moving from nose asserts to testcase asserts seems like a real loss of time for both of us.

jonkiparsky:
As far as I can tell, this is good to merge. @gawaineo any objections?

bethlakshmi:
Merge conflict was fixed by me.  reran unit tests.

PR #837, opened 2016-07-17
[GBE-665] Send a notification when volunteer schedule changes

jonkiparsky:
Have not tested on test server, but it looks good from here.


expo/gbe/models.py
```diff
@@ -344,6 +344,10 @@ def schedule(self):
             resources_allocated__resource__worker___item=self)]
         return sorted(set(events), key=lambda event: event.start_time)
 
+    def volunteer_schedule(self, conference=None):
+        conference = conference or Conference.current_conf()
```
bethlakshmi: After seeing how many times in a thread we look up the current conference in a single request/response with the debug toolbar, I'm worried about the pattern.  We reuse it all the time, but I'd like us to consider picking where we choose the conference and then streamlining all the code so that the conference selection only happens once in a thread.  Not a blocker.

expo/gbe/views/create_volunteer_view.py
```diff
@@ -77,3 +70,13 @@ def CreateVolunteerView(request):
                        'page_title': page_title,
                        'view_title': view_title,
                        'nodraft': 'Submit'})
+
+
+def notify_volunteer_reviewers(user_profile):
```
bethlakshmi: If we're going to abstract this, could we bring it higher than the volunteer view level?  Let the view give template name, bidder name, bid type and URL and then use a common function.  Long term, I can see the value to other reviewers getting similar mail for the other bid types, so while we're refactoring, let's make the abstraction level higher.

expo/gbe/functions.py
```diff
@@ -67,6 +68,23 @@ def mail_to_group(subject, message, group_name):
     return None
 
 
+def mail_schedule_to_user(subject, user, start_message, end_message):
```
jonkiparsky: Um... I'm not sure why you're adding this here.

expo/gbe/models.py
bethlakshmi: I worry about adding to the auto-selection of conference at the model level, because when I started looking at the django-debug-toolbar -we had conference selection happening all over the place.
I'm not ready to call it a blocker, because I haven't dug in far enough on the analysis of where and how it's happening... but I'd like us to figure out at what level we detect the target conference, and then do it in one place consistently.

expo/scheduler/models.py
```diff
@@ -449,6 +449,11 @@ def get_conflicts(self, new_event):
                 conflicts += [event]
         return conflicts
 
+    def volunteer_events(self, conference):
```
bethlakshmi: this seems similar but different from get_bookings.  Could you combine them?  Filtering by conference (if present) sounds great.  And if you're getting only the volunteer events, then they should be only for the role="Volunteer" - I'd prefer to have fewer functions here, so that if we ever have a bug to fix, or a logic change, it's in common code.

expo/scheduler/models.py
```diff
@@ -481,52 +486,6 @@ def __unicode__(self):
         except:
             return "No Worker Item"
 
-'''
-class EquipmentItem(ResourceItem):
```
bethlakshmi: Yay!  Good bye unused code!

expo/scheduler/templates/scheduler/tag_templates/volunteer_schedule.tmpl
```diff
@@ -0,0 +1,9 @@
+{% if schedule %}
```
bethlakshmi: Is there any reason to call this a "volunteer schedule"?  It's really just a schedule...  reusable for all sorts of great things.

expo/scheduler/views.py
```diff
@@ -440,6 +443,18 @@ def show_potential_workers(opp):
...
+    subject = "A change has been made to your Volunteer Schedule!"
+    message = loader.get_template('scheduler/volunteer_schedule_update.tmpl')
+    c = Context({'profile': profile})
+    if not settings.DEBUG:
+        mail_to_user(subject, message.render(c), profile.user_object)
+
+
 @login_required
 def allocate_workers(request, opp_id):
     '''
```
bethlakshmi: The same notification call needs to be made in /expo/gbe/views/volunteer_changestate_view.py - the flow for approving a volunteer application.  I think you can just add the call to notify_volunteer_schedule_change there.

expo/gbe/functions.py
bethlakshmi: me either... I didn't commit this intentionally - as soon as I synced with GBE-665 this showed up.

bethlakshmi:
**Must have:**
- call the notification email in /expo/gbe/views/volunteer_changestate_view.py  so the functionality is consistent in the two ways of allocating volunteers.
- make volunteer_events specific to being a "volunteer" - or make the same call that we make when we display the schedule on the landing page (all things the profile, personae and acts the personae are in that have bookings)
- fix (or show me how to fix) the problem I'm having in unit testing (posting in a sec)


**Very strong nice to have:**
- combine get_bookings and volunteer_events 

The others would make me happy, but I could live w/out the joy.

expo/tests/scheduler/test_allocate_workers.py
```diff
@@ -1,3 +1,4 @@
+from django.core import mail
```
bethlakshmi: I get 4 failures.  See attached.
[scheduler_tags error.txt](https://github.com/bethlakshmi/GBE2/files/372918/scheduler_tags.error.txt)

expo/gbe/models.py
jonkiparsky: I think this is a good issue to address, but not one I want to address here. You're right, this pattern has got a bit out of control and we should figure out a way to address it.

expo/scheduler/templates/scheduler/tag_templates/volunteer_schedule.tmpl
jonkiparsky: Good point.

expo/gbe/views/create_volunteer_view.py
jonkiparsky: This sounds premature to me. Let's do this when we need it.

expo/gbe/models/remains.py
```diff
@@ -1931,3 +1950,7 @@ class ProfilePreferences(models.Model):
     class Meta:
         verbose_name_plural = 'profile preferences'
         app_label = "gbe"
+
+
+def mail_to_user(subject, message, user):
```
bethlakshmi: this is a basic function call that has nothing to do with models.  Why put it in a model py file?

expo/scheduler/models.py
```diff
@@ -425,6 +425,9 @@ def get_bookings(self, role='All'):
             events = Event.objects.filter(
                 resources_allocated__resource__worker___item=self,
                 resources_allocated__resource__worker__role=role)
+        if conference:
+
+            events = events.filter(eventitem__event__conference=conference)
```
bethlakshmi: huzzah!
I love this

expo/tests/scheduler/test_allocate_workers.py
```diff
@@ -269,3 +270,15 @@ def test_post_form_valid_delete_allocation_w_bad_data(self):
             response,
             '<form method="POST" action="/scheduler/allocate/' +
             str(self.volunteer_opp.pk) + '"', count=1)
+
```
bethlakshmi: There should be something similar that covers each case - it looks to me like there are 3 basic cases:
- a volunteer's assignments are edited via the volunteer change state (ie the assign volunteer form, any submit will do it if it's valid)
- a volunteer is assigned via the volunteer opportunity edit (the scheduler view when the allocation is added)
- a volunteer's obligation is deleted.

In this test it looks like only 1 case is exercised.

bethlakshmi:
I could live with an answer to the question on models.py having what is purely an email function in it.
But I'd really really like to see the 3 test cases I mentioned covered.

Do that and I'm glad to approve.

jonkiparsky:
Yeah, that function shouldn't be where it is. Circularities prevented me putting it in functions, which is where I wanted it to be. 
I would be happy to attack that problem if you want - but the resulting pull request will touch a lot of code.

bethlakshmi:
On the models function - meh - I have my answer I can live with it....
Can you give some more tests?

BenignCremator:
Read code, and current conversation, which largely answered the questions I had.  Willing to approve.

expo/tests/scheduler/test_allocate_workers.py
jonkiparsky: Added tests

PR #846, opened 2016-07-30
GBE 822:  rankable, editable list

bethlakshmi:
- the editable list for a volunteer bid has been updated for:
   - the original bid
   - editing the bid
   - reviewing the bid as the bid committee
   - viewing the bid passively as as the submitter.

- this also needed bid-related updates to:
   - the review volunteer list 
   - assign volunteer 

- also the scheduled volunteer opportunity offers a connection to the same set of interest values, so that needed update on manage volunteers, and volunteer interest searched.

In testing I also found that inadequate conference windows can really mess up the volunteer bid, so I did checks and created an error message when we don’t have windows or interests ready.

And I broke up the assist_volunteer way complicated test (1 test) into 5-6 other, simpler tests.

General cleanup
- code coverage should be the same or better - added tests when I was working in a given view
- pep8 fix is the only thing in the ticketing function 
- added volunteer_display_functions to reduce deprecated code in review and view 

and my first south data migration


expo/gbe/cms_app.py
```diff
@@ -2,9 +2,10 @@
 from cms.apphook_pool import apphook_pool
 from django.utils.translation import ugettext_lazy as _
 
-#this is largely a proof of concept right now, I haven't found this this was useful.
+
+# this is largely a proof of concept right now,
```
jonkiparsky: Is this serving any purpose? The comment indicates that it isn't, in which case it should be deleted.

expo/gbe/migrations/0009_auto__add_availableinterest__add_volunteerinterest__add_unique_volunte.py
```diff
@@ -0,0 +1,421 @@
...
+        db.create_unique(u'gbe_volunteerinterest', ['interest_id', 'volunteer_id'])
+
+        # Adding field 'GenericEvent.volunteer_type'
+        db.add_column(u'gbe_genericevent', 'volunteer_type',
+                      self.gf('django.db.models.fields.related.ForeignKey')(to=orm['gbe.AvailableInterest'], null=True, blank=True),
+                      keep_default=False)
+
+
+    def backwards(self, orm):
+        # Removing unique constraint on 'VolunteerInterest', fields ['interest', 'volunteer']
```
jonkiparsky: These comments are gratuitous.

expo/gbe/models/available_interest.py
```diff
@@ -0,0 +1,16 @@
+from django.db import models
+
+
+class AvailableInterest(models.Model):
+    interest = models.CharField(
+        max_length=128,
+        unique=True)
+    visible = models.BooleanField(default=True)
+    help_text = models.TextField(max_length=500,
```
jonkiparsky: Is max_length specified for a reason? (not enforced on a TextField)

expo/gbe/views/create_volunteer_view.py
```diff
@@ -9,8 +12,12 @@
 from django.shortcuts import render
 
 from expo.gbe_logging import log_func
-from gbe.forms import VolunteerBidForm
+from gbe.forms import (
+    VolunteerBidForm,
+    VolunteerInterestForm
```
jonkiparsky: trailing comma, please

expo/gbe/views/create_volunteer_view.py
```diff
@@ -1,6 +1,9 @@
 from django.contrib.auth.decorators import login_required
 from django.contrib import messages
-from django.http import HttpResponseRedirect
+from django.http import (
+    HttpResponseRedirect,
+    Http404
```
jonkiparsky: trailing comma, please

expo/gbe/views/create_volunteer_view.py
```diff
@@ -19,31 +26,65 @@
     validate_profile,
 )
 from gbetext import (
-    default_volunteer_submit_msg
+    default_volunteer_submit_msg,
+    default_volunteer_no_interest_msg,
+    default_volunteer_no_bid_msg
```
jonkiparsky: again, trailing comma

expo/gbe/views/create_volunteer_view.py
```diff
@@ -19,31 +26,65 @@
...
+            interest_form = VolunteerInterestForm(
+                request.POST,
+                initial={'interest': interest},
+                prefix=str(interest.pk))
+            formset += [interest_form]
+            if interest_form.is_valid():
+                if int(interest_form.cleaned_data.get('rank')) > 3:
+                    like_one_thing = True
+            else:
+                valid_interests = False
```
jonkiparsky: This could be a lot cleaner. With minimal effort it could be extracted to a function. However, it could also be reduced to a line or two, something like this:

```
formset = filter(lambda f: f.isvalid(), [create_form(interest) for interest in available_interests])
if form.is_valid() and formset and any([like(f) for f in formset]):
  etc.
```
where create_form initializes a VolunteerInterestForm and like() figures out whether there is a positive preference expressed here.
In this case, there wouldn't really be a need to extract the one line to a helper function.

expo/gbe/views/create_volunteer_view.py
```diff
@@ -54,6 +95,10 @@ def CreateVolunteerView(request):
                 for window in form.cleaned_data['unavailable_windows']:
                     volunteer.unavailable_windows.add(window)
 
+                for interest_form in formset:
+                    vol_interest = interest_form.save(commit=False)
```
jonkiparsky: What if one or more forms in formset are not valid?

expo/gbe/views/create_volunteer_view.py
```diff
@@ -70,24 +115,38 @@ def CreateVolunteerView(request):
                 messages.success(request, user_message[0].description)
             return HttpResponseRedirect(reverse('home', urlconf='gbe.urls'))
         else:
+            formset += [form]
+            if not like_one_thing:
+                user_message = UserMessage.objects.get_or_create(
+                    view='CreateVolunteerView',
+                    code="NO_INTERESTS_SUBMITTED",
+                    defaults={
+                        'summary': "Volunteer Has No Interests",
```
jonkiparsky: We'll have to figure out a reasonable way to handle these literal strings, but not a blocker at the moment.

expo/gbe/views/edit_volunteer_view.py
```diff
@@ -7,15 +7,20 @@
 from django.http import HttpResponseRedirect
 from django.core.urlresolvers import reverse
 from expo.gbe_logging import log_func
-from gbe.forms import VolunteerBidForm
+from gbe.forms import (
+    VolunteerBidForm,
+    VolunteerInterestForm
```
jonkiparsky: trailing comma...

expo/gbe/views/edit_volunteer_view.py
```diff
@@ -33,44 +39,81 @@ def EditVolunteerView(request, volunteer_id):
             available_windows=the_bid.conference.windows(),
             unavailable_windows=the_bid.conference.windows())
 
-        if form.is_valid():
+        valid_interests = True
```
jonkiparsky: this looks like duplicated code.

expo/gbe/views/edit_volunteer_view.py
```diff
@@ -33,44 +39,81 @@ def EditVolunteerView(request, volunteer_id):
...
     else:
-        if len(the_bid.interests.strip()) > 0:
-            interests_initial = eval(the_bid.interests)
-        else:
-            interests_initial = []
-        form = VolunteerBidForm(
+        # originally the volunteer create was supposed to save a title, there
+        # was a bug and most old bids don't have a title.
+        # This is the autocorrect
+        if not the_bid.title:
```
jonkiparsky: would it be better to do this in a migration instead?

expo/tests/fixtures/tickets_to_events.json
```diff
@@ -1,1367 +0,0 @@
-[
```
jonkiparsky: Hurrah!

jonkiparsky:
In general, I'm having a really hard time with the way these tests are developing, but I'm willing to let this pass for now. However, there are some important style issues and a potential bug that need to be cleaned up. (already commented)

expo/gbe/cms_app.py
bethlakshmi: The only reason this was touched was pep8 fixing.
The code is not in use on the site, but it IS a useful stub if we address the desire to put forms within CMS pages - this is the foundation of that, but it was never finished.

expo/gbe/models/available_interest.py
bethlakshmi: I try to be in the habit of always limiting length.  Didn't realize textfield doesn't use it - but I see that in the DB.  will pull it.

expo/gbe/views/create_volunteer_view.py
bethlakshmi: I'm not trying to end up with a formset filtered on validity here - the formset needs to include the collection of **all** interest forms or it will not render correctly in the case of an error because the invalid form will be missing from the formset.

And I need to know if any 1 form in the formset is invalid.  The way the second line you wrote is expressed, if any form in the form set is valid, you'll be good to go - I need the opposite.

I'll think on this more, but it seems counter productive to me to go through the same loop numerous times.  I think I end up writing 3 for loops for your method:
1 - loop to build the formset (no filter, every form, valid or invalid)
2 - loop to figure out if there are any invalid forms
3 - loop of valid forms to be sure that there is at least one selected interest

that takes it from 1 loop to 3, causing more iterations for no reason.

jonkiparsky: Worst case, the three loops do the same work as the one loop, so nothing is lost. More likely, python is able to optimize the loops in the filter and the any, so we save some work. In either case, the code is infinitely more readable which is always the first thing to shoot for. (Knuth's law)
Feel free to use an `all` instead of an `any` - this was a quick example, not drop-in code.

jonkiparsky: Or, if you really don't want to make this readable, you can extract it to a function with a name that tells me what it's doing, and I'll make it readable the next time I touch this code.

expo/gbe/views/create_volunteer_view.py
bethlakshmi: They are checked for validity on lines 75 through 85 and the only way to get to this line is line 86 - an if statement checking that validity.

BenignCremator:
Read diffs, no comments not already made.

expo/gbe/cms_app.py
jonkiparsky: I don't much like leaving dead code lying around, but okay, leave it.

expo/gbe/views/create_volunteer_view.py
jonkiparsky: Reviewing the code, I see that we can't get here with a bad form in the formset. But the fact that this was hard does indicate that readability is an issue here. Extraction and naming of the offending code is fine, we'll fix it next time it gets touched.

expo/gbe/views/edit_volunteer_view.py
bethlakshmi: I dig it.  I'll work on fixing that, and I *think* that's my last update on this before I request a re-review.

PR #863, opened 2016-08-22
GBE-305: Now showing persona pics

bethlakshmi:
- refactored “promo_small” to be promo_mini and promo_thumb for the two different sizes
- added the code to make the picture appear when it’s there, and yet justify
nicely when it’s not.
- added a test to landing_page tests

### DIFF
--- a/expo/gbe/models/remains.py
+++ b/expo/gbe/models/remains.py
@@ -483,19 +483,21 @@ class Performer (WorkerItem):
         '''
         return alerts
 
-    '''
-    gets the URL for images that have been sized specifically to
-    suit GBE layouts
-       mini = the size for show/act and fashion fair grids - height = 125p
-       thumb = the size for the /gbe page -  height 75-100p
-    There is a timebased script that minifies uploaded images.  When using
-    this function, always check first that the image exists, and if it
-    does not, default to the profile_image.name
-    '''
     def get_promo_image_sizes(self, size):
-        pieces = self.promo_image.name.split('/')
-        pieces.insert(-1, size)
-        return '/'.join(pieces)
+        '''
+        gets the URL for images that have been sized specifically to
+        suit GBE layouts
+            mini = the size for show/act and fashion fair grids - height = 125p
+            thumb = the size for the /gbe page -  height 75-100p
+        There is a timebased script that minifies uploaded images in these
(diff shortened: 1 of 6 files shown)


expo/gbe/models/remains.py
```diff
@@ -483,13 +483,29 @@ def append_alerts(self, alerts):
...
 
-    @property
-    def promo_small(self):
+    '''
+    gets the URL for images that have been sized specifically to
+    suit GBE layouts
+       mini = the size for show/act and fashion fair grids - height = 125p
+       thumb = the size for the /gbe page -  height 75-100p
+    There is a timebased script that minifies uploaded images.  When using
+    this function, always check first that the image exists, and if it
```
jonkiparsky: Shouldn't this function be responsible for this check, if it's needed every time it's called?

expo/gbe/models/remains.py
```diff
@@ -483,13 +483,29 @@ def append_alerts(self, alerts):
         '''
         return alerts
 
-    @property
-    def promo_small(self):
+    '''
```
jonkiparsky: Docstring should be the first thing evaluated in the function - not preceding the function definition.

expo/gbe/models/remains.py
bethlakshmi: Sure.. but the template will still need to check it.  I'll add a check here to return None to make for cleaner code.

BenignCremator:
Tested functionality, looks good.  On reading code, one thing that jumped out at me was that there were multiple, identical blocks begining "{% if performer.promo_thumb %}".  We may want to pull these out into their own templates to make it easier to use this elsewhere.  Beyond that, looked good to me.  I am giving it approval.  How many approvals before pulling?

jonkiparsky:
I totally approve of Hunter's suggestion. It shouldn't be too hard to make that change - @bethlakshmi are you up for it?
If not, not a blocker, but it would be a really good improvement.

bethlakshmi:
do guys mind if i just pull when i'm done w/ @BenignCremator's suggestion? 

I like the refactor and will gladly do either way - I am just eager to get this one gone...

bethlakshmi:
Taking the initiative and pulling this...

PR #868, opened 2016-08-24
GBE-866: Prevent server error if user has two existing acts with name conflicts

jonkiparsky:
This shouldn't happen, but if someone does have two acts with conflicting names the get() call fails hard. This is not a great solution - it redirects them to one of the two conflicting acts - but at least it doesn't die in flames.


bethlakshmi:
ick.

I approve this.  Second approver is free to pull if he likes it.

BenignCremator:
Tested functionality of the fix.  Seems to work well.  That said, it works as described.  Trying to submit act 675 takes one to the edit form for act 674.  Would prefer that the code selects that one that matches the act id in the request url then the first one from the DB, but that is a fight for another day.  Reading the code shows two lines of change, and they were quite comprehensible.  Merging and closing.

PR #862, opened 2016-08-21
Gbe 748: Refactor review_X_view and view_X_view to class-based views

jonkiparsky:
Added new superclass views ReviewBidView and ViewBidView, and made ReviewActView, ReviewClassView, etc, and ViewActView, ViewClassView, etc, subclasses of those. 

Info here: https://docs.djangoproject.com/en/1.10/topics/class-based-views/

PR #869, opened 2016-08-25
GBE-867: fix lack of import on act change state

bethlakshmi:
This came up when I was testing PR #862, but it is really it’s own problem.  When Act Change State was refactored from views.py into it’s own file, i believe we lose the import for django.contrib.messages.

Added the import and a test so it won’t happen again.


gawaineo:
Reviewed and tested. Everything looks good. A potential code cleanup (not related to your changes but in the same file) change my editor pointed out is on line #3 in the import statements `(from django.core.exceptions import PermissionDenied)` is not being used anywhere.

jonkiparsky:
@gawaineo Good catch, fixing that.

jonkiparsky:
Oh, oops. Thought that was one of mine! Anyway, I'll push the fix. Sorry @bethlakshmi

jonkiparsky:
Anyway, this looks fine to me. Merging.

PR #870, opened 2016-08-25
Gbe 858 - Some bug fixes, minor improvements and tweaks to the Vagrantfile and dbreset

BenignCremator:
Mostly minor changes.  Take a look and let me know what you think.


bethlakshmi:
Reprovisioned my vagrant, setup a new db and did minor tests on the local server and the unit tests - all works well.  I'm glad to approve.

Vagrantfile
```diff
@@ -56,31 +54,42 @@ $bootstrap = <<BOOTSTRAP
...
   chown -R vagrant:vagrant /home/vagrant
-  echo "source /home/vagrant/aliases" >> /home/vagrant/.bashrc
-  ifmkdir /vagrant/expo/logs; ifmkdir /vagrant/uploads
-  ifmkdir /vagrant/uploads/images; ifmkdir /vagrant/uploads/images/fullsize
-  ifmkdir /vagrant/uploads/images/mini; ifmkdir /vagrant/uploads/images/thumb
-  chown -R vagrant:vagrant /vagrant
+  echo "source /home/vagrant/aliases" >> /home/vagrant/.bashrc 
+
+###  The next four lines will have to be done by hand, until we get a good test DB setup 
+###  cat /vagrant/config/gbe2016test_data.sql | python manage.py dbshell
```
jonkiparsky: You can also do 
python manage.py dbshell < YOUR_FAVORITE_DB_DUMP.sql

Vagrantfile
```diff
@@ -56,31 +54,42 @@ $bootstrap = <<BOOTSTRAP
...
+###  python manage.py syncdb --all
+###  python manage.py migrate --all
+###  python manage.py collectstatic
+
+  if [ -f /vagrant/config/local_settings.py ]
+  then
+      cp /vagrant/config/local_settings.py /vagrant/expo/expo
+  fi
+  if [ -f /vagrant/config/gbe_*.sql ]
```
jonkiparsky: This seems to assume a naming convention that is not like the one you use on the live server, meaning we have to remember to rename database dumps when we take them down from the server. Could we get these in sync?

Vagrantfile
```diff
@@ -27,8 +26,7 @@ $bootstrap = <<BOOTSTRAP
   sudo chown -R vagrant /var/lib/
   sudo chown -R mysql /var/log/mysql
   sudo chown -R mysql /var/log/mysql
-  sudo apt-fast -y install emacs24-nox
-  sudo apt-fast -y install libmagickwand-dev
+  sudo apt-fast -y install emacs24-nox findutils realpath libmagickwand-dev
```
jonkiparsky: emacs is, strictly speaking, not required for our build. Can we come up with a way to make it optional, so that there's less build time for people who don't want it on their vagrant box?

jonkiparsky: What happens if one of these installs fails? Do the rest go through, or does the whole line fail, or do we exit on failure, or what?

jonkiparsky:
*BLOCKER* Since this moves us over to dbreset for setting up our test machines, I'm a not willing to adopt this until we're sure we've isolated the problems that @BenignCremator has been having with his DB config/test running and found that they're not part of the dbreset process. (or it they are, that we've solved them and can consistently build a working db with that process)

Vagrantfile
BenignCremator: On 09/01/2016 10:42 PM, Jon Kiparsky wrote:

> This seems to assume a naming convention that is not like the one you
> use on the live server, meaning we have to remember to rename database
> dumps when we take them down from the server. Could we get these in sync?

Will do.  Moving it into a var so it can be more easily changed in just 
one place.

Hunter

Vagrantfile
BenignCremator: On 09/01/2016 10:46 PM, Jon Kiparsky wrote:

> emacs is, strictly speaking, not required for our build. Can we come up
> with a way to make it optional, so that there's less build time for
> people who don't want it on their vagrant box?

Yes....  But not right now.  Pulling it for the time being, and waiting 
until I start looking at some of the other improvements I would want to 
do on the Vagrantfile....  Like having an external port setting so two 
box do not fight for 8282.

Hunter

BenignCremator: On 09/01/2016 10:47 PM, Jon Kiparsky wrote:
> In Vagrantfile
> <https://github.com/bethlakshmi/GBE2/pull/870#discussion_r77286459>:
>
>> @@ -27,8 +26,7 @@ $bootstrap = <<BOOTSTRAP
>>    sudo chown -R vagrant /var/lib/
>>    sudo chown -R mysql /var/log/mysql
>>    sudo chown -R mysql /var/log/mysql
>> -  sudo apt-fast -y install emacs24-nox
>> -  sudo apt-fast -y install libmagickwand-dev
>> +  sudo apt-fast -y install emacs24-nox findutils realpath libmagickwand-dev
>
> What happens if one of these installs fails? Do the rest go through, or
> does the whole line fail, or do we exit on failure, or what?

The rest go through.  I replaced:

   sudo apt-fast -y install emacs24-nox

with:

   sudo apt-fast -y install argle-bargle

on an unbuilt Vagrantfile, up'ed it, and it spewed red errors about the 
package not found, and continued.

After that, I cleared argle-bargle, and replaced it with:

   sudo apt-fast -y install realpath looseutils libmagickwand-dev

because if there is a utility package to help you find files, there 
should also be one that helps you loose them, but there is not.  Virtual 
Box and Vagrant failed to install realpath and libmagickwand-dev, but 
only reported not being able to find looseutils.  Surprising, I was 
expecting more apt-get like behavior.  Separating them over three lines. 
  Good catch.

Hunter

BenignCremator:
On 09/01/2016 10:51 PM, Jon Kiparsky wrote:

> /BLOCKER/ Since this moves us over to dbreset for setting up our test
> machines, I'm a not willing to adopt this until we're sure we've
> isolated the problems that @BenignCremator
> <https://github.com/BenignCremator> has been having with his DB
> config/test running and found that they're not part of the dbreset
> process. (or it they are, that we've solved them and can consistently
> build a working db with that process)

Expected.  Spending tonight building 5 separate vboxes to test each 
possible intersection on that:

1:  no dbreset and no db file in config, have py.test --create-db based
     purely on the models.py files.
2:  no dbreset but with cleansed db dump that alters URL to localhost
3:  dbreset and cleansed db file
4:  dbreset and plane db file (URLs point to production server)
5:  no dbreset and plane db file

Will report once all of that is done.

Hunter

PR #871, opened 2016-08-26
Gbe 389 - Working DjangoBB-Forums with way to turn it off in local-settings.py

BenignCremator:
Working DjangoBB-Forums, in that you can create categories, forums, post, and profiles, and nothing seems to break in GBE.  Need to redirect the forum profiles back to GBE profiles, get user creation to automatically create a forum profile, and a few other things.


bethlakshmi:
I'm having a hard time reviewing this - the 293 different files is making it really hard to compare changes.  And it seems like the ignore files list is quite different, because I have 253 files in my local that are now tracked as uncommitted changes.

I think if you sync from master to your branch, this might simplify the pull request tremendously - @dracus - could you give that a try and I'll head in to review again?

BenignCremator:
Ok, too much water under the bridge, it is easier to close this, delete the branch, recreate the branch from master, redo the work, and reissue to PR.

PR #875, opened 2016-08-27
GBE-872: Update requirements.txt

jonkiparsky:
Do the usual fandango:

syncdb or load an existing db image, run migrations, run tests with --create-db flag first time

### DIFF
--- a/config/requirements.txt
+++ b/config/requirements.txt
@@ -1,7 +1,7 @@
-Django==1.6.11
+django==1.6.11
 MySQL-python==1.2.3
-South
-Pillow
+South==1.0.2
+Pillow==3.3.0
 cmsplugin-bootstrap-carousel==0.2.2
 cmsplugin-filer==1.0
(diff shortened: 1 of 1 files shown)


bethlakshmi:
Weird - when I rebuilt (and restored my self-destroyed local settings) then all unit tests pass.  But my server port had changed (was 8282 now, 2200).  It's hard for me to see how that happened.  There's been enough virtual machine shift and my own Mac went through an upgrade - so it could be external influences... but .. weird...

I'm game to approve it - just putting out the marker for reviewer 2

I'll keep using these settings to make sure nothing weird happens.

jonkiparsky:
This implies that you had a machine taking to port 8282.

sent from my handy. typos are just part of the fun.

bethlakshmi:
fair enough - we'd been doing enough different vagrant related changes recently that I started getting nervous.  I'm glad to approve - whoever hops on board after me is free to push the button. :)

gawaineo:
Review and tested. All tests passed from runcoverage and py.test --create-db

PR #880, opened 2016-08-29
GBE-255: add an ad-rotator

bethlakshmi:
- needs a reprovision of vagrant.
- after reprovision and re-setup, do a migrate

you can see the work manually by:
- go to localhost:####
- use a user that can edit django-cms (scratch and betty can)
- in the top white bar, click edit
- click structure
- insert a plugin in any box (it’s inserted through the menu in the upper right hand corner of the box.
- plugin is in the GBE area and called Ad Rotator.

——
Then go to /admin
- see that there is now “Ad_Rotator”
- and in it “Banner ads”
- you have to add banner ad items with start/end dates before and after now.

——
When you’ve set up both a page with the banner ad, and the images themselves, the ad picture will appear on refresh

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,3 +108,5 @@ htmlcov
 expo/.coverage
 
 .cache/v/cache/lastfailed
+
+expo/src/cmsplugin-image-gallery-master/
--- a/config/requirements.txt
+++ b/config/requirements.txt
@@ -50,3 +50,4 @@ ipython
 pep8
 django-debug-toolbar==1.3.2
 wand
+django-ad-rotator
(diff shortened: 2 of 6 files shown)


BenignCremator:
Functionality tests look good.  Very basic, needs user to scale image by hands, and does not rotate them in place, but on every page load.  But it works.  Moving on to coverage and code review.

bethlakshmi:
yep.  it is the ad-rotator @burlexpo asked for.  I decided that since we went with the more minimal option, I would not assume any more requirements than what were given.  It's actually possible to force-scale the template, but I decided not to muck with that on this ticket for fear of scope creep.

BenignCremator:
Code review looks great.  Coverage is not working in my environment, with foreign key problems.  I approve of the PR as it, and if someone else can get py.test to work, this PR should be merged.

BenignCremator:
Gawaine test and reviewed this, and gave his approval over Slack.  Merging and closing.

PR #898, opened 2016-10-02
GBE-110: 2 votes, 2 shows for acts

bethlakshmi:
- fixed the bug making it impossible to edit and update an eval
- 3 more tests
- pep8
- removed code not covered or used in any way I could see.
- refactored query loading so it’s loaded with the form and not at init, which keeps the list of current shows from being statically initialized once.


jonkiparsky:
This seems to clear up the functional issue that Betty.  I see we're still creating new ShowVote objects, but I'm okay leaving that in order to get this closed. 

Approved for merge. @BenignCremator  or @gawaineo please review and merge if acceptable.

BenignCremator:
All looks good, approving and merging.

PR #910, opened 2016-11-28
GBE-908:  Selected caching disabled

bethlakshmi:
I’ve disabled caching from:

- all edit bid pages
- all review bid list display pages
- all voting and change state action pages (so that you can see what you voted or changed most recently when reloading)
- profile management - both for admin and regular users
- the landing page for a logged in user
- volunteer assignment
- scheduler views - the privileged areas - event edit, the event editing list displays, act scheduling, contact email lists 
- reports that are focused on print runs, vs. regular reviews

The following will still have a 10 minute cache:
- pages that are primarily for static public view (event details, event view-only lists, fashion fair, show/act displays)
- pages with no user-made data (submit for the first time pages - like create bids, create profiles, etc)
- pages that are static that staff download way more than they change (some of the reports).

The marker used consistently is @never_cache, which is documented here:

https://docs.djangoproject.com/en/1.7/topics/cache/#the-per-site-cache

For a test:
- make sure you’ve got no cache setup in your local (DummyCache turns off all caching)
- use Review Acts
- cast a vote
- verify that the vote is updated on the review act list
- verify that the vote is updated on the “review” link for the act you submitted.

That’s an example, for each @never_cache you could do the same kind of test for each thread.

I did that manually, as caching doesn’t seem to apply to unit tests.

Unit tests and pepdiff remain in a good state.

### DIFF
--- a/expo/gbe/report_views.py
+++ b/expo/gbe/report_views.py
@@ -4,6 +4,7 @@ from django.http import HttpResponse, HttpResponseRedirect, Http404
 from django.core.urlresolvers import reverse
 from django.db.models import Q
 from django.core.management import call_command
+from django.views.decorators.cache import never_cache
 
 import gbe.models as conf
 import scheduler.models as sched
@@ -89,6 +90,7 @@ def staff_area(request, area_id):
                    'area': area})
 
 
+@never_cache
 def env_stuff(request, conference_choice=None):
     '''
     Generates an envelope-stuffing report.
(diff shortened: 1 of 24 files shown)


gawaineo:
Everything looks good, one minor thing to fix (if even)....pep8 on line 503, pepdiff snippet below:

```
vagrant@blueprint:/vagrant/expo$ pepdiff
/vagrant /vagrant/expo /vagrant/expo
expo/gbe/report_views.py:503:80: E501 line too long (92 > 79 characters)
```

P.S. Will approve after pep8 update.

Issue #593, opened 2015-12-05
Eliminate lookup by show name in act tech info edit flow


jonkiparsky:
The act tech info edit view  (/acttechinfo/edit/389) does a lookup by show name, which fails if show name is duplicated. This lookup should be done by show ID, so that show names are not forced to be unique. 

See issue #587 for the details of the server error which revealed this issue.

bethlakshmi:
show = get_object_or_404(
                Show,
                title=request.POST['show']).scheduler_events.first()

bethlakshmi:
sigh

jonkiparsky:
see also #353 for similar problem

BenignCremator:
BLOCKER:  Went to URL listed in original ticket, altered some data (changed a backlight color from white to OFF), and pressed submit.  Error when submitting an edit:
MultipleObjectsReturned at /acttechinfo/edit/389
get() returned more than one Show -- it returned 2!

Is this is the same error that this ticket covers in a different but dependent location?  Either way, submit is broken, and I can not approve until there is basic functionality in standard cases.

PR #914, opened 2016-12-02
GBE-593: fix show title bug in act scheduling

### LINKED ISSUE #353: Can't use () or . in Show due to schedule_acts URL lookup
I had a test show called "The Bordello (Fri. Late)" in my test database.

It failed on a URL reverse lookup until I changed it to "The Bordello".

I'd much rather we use ids than names for any URL lookup.  We do absolutely nothing to guarantee names are unique, so looking up names will always produce non-deterministic results.

bethlakshmi:
This is the bug I stumbled over last night.  #593 

It also closes #353.

When you navigate to the act scheduler on a show that has the same title as another show, you got an error because we filtered on title, not on private key.

This fixes that.

Example from current test data:

http://gbe2016test.gbeadmin.webfactional.com/scheduler/edit/Show/2518

click on- schedule acts:

http://gbe2016test.gbeadmin.webfactional.com/scheduler/acts/The%20Main%20Event

Note the URL, and you can see the fail on test.  On this build, the URL will be an id, not a title, and the thread will work.

### DIFF
--- a/expo/expo/settings.py
+++ b/expo/expo/settings.py
@@ -259,6 +259,38 @@ USE_L10N = True
 
 USE_TZ = True
 
+# TIME_FORMAT is often used for ending times of events, or when you
+# do not need to give the date, such as in calendar grids
+try:
+    TIME_FORMAT
+except:
+    TIME_FORMAT = "%-I:%M %p"
+
+try:
+    DATE_FORMAT
+except:
+    DATE_FORMAT = "%a, %b %-d"
+try:
+    DATETIME_FORMAT
+except:
+    # Default DATETIME_FORMAT - see 'man date' for format options
+    # DATETIME_FORMAT = "%I:%M %p"
+    DATETIME_FORMAT = DATE_FORMAT+" "+TIME_FORMAT
+
+try:
+    SHORT_DATETIME_FORMAT
+except:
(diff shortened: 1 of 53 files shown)


BenignCremator:
Code review and py.tests look good, moving on to functional test.

BenignCremator:
BLOCKER: Went to URL listed in original ticket, altered some data (changed a backlight color from white to OFF), and pressed submit. Error when submitting an edit:
MultipleObjectsReturned at /acttechinfo/edit/389
get() returned more than one Show -- it returned 2!

Is this is the same error that this ticket covers in a different but dependent location? Either way, submit is broken, and I can not approve until there is basic functionality in standard cases.

bethlakshmi:
Sweet!  Great find! 

Will fix!

Thank so, so much for catching that!  I came at the code a different way and didn't even realize I had missed the original target problem.  Sigh.

gawaineo:
Ran py.test, test results look good. Verified page visually, the event_id is being used instead of the event/show name. Everything looks good to me, I approve.

bethlakshmi:
OK, @BenignCremator - I think I covered the original bug, as well, now.
Please re review

BenignCremator:
Much better.  Functionality works, submission works, py.test is happy, and code review looks good.  Approving and merging.

PR #915, opened 2016-12-04
GBE-679 cleanup #1: just scheduler views

bethlakshmi:
@BenignCremator and @gawaineo 

please try this on your machines that have a problem with GBE-679 and let’s see if it passes.

This is a small fragment of the GBE-679 cleanup work.  I figured we might be able to pull piece by piece and see where there’s a problem.

I think this is unlikely to be part of the problem, because it’s no where near Volunteer.

### DIFF
--- a/expo/scheduler/views/__init__.py
+++ b/expo/scheduler/views/__init__.py
@@ -1,0 +1,27 @@
+from remains import (
+    get_events_display_info,
+    get_event_display_info,
+    event_list,
+    detail_view,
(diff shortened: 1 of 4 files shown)


gawaineo:
@bethlakshmi @BenignCremator ran py.test all tests passed. Went through the application as well, everything looks good to me. I approve.

BenignCremator:
Functionality and py.test both work well.  Approving and merging.

PR #921, opened 2016-12-10
GBE-917: add contact by role links to special menu

bethlakshmi:
When I did that, it meant that the constraints of the existing design (1 menu item per role) no longer work.  I redid the structure of the menu format to let us have:
- a hierarchy for easier readability
- a many to many mapping of menu item to role.


BenignCremator:
Talked to Bethlakshmi about the questions I had.  Menuing has moved into DjangoCMS, so the templating menus was not in use.  Approving and merging.

PR #925, opened 2016-12-11
GBE-924:  Fix for GBE-679 cleanup #1 -- URGENT

bethlakshmi:
I had to manually merge the scheduler area to get GBE-679 to work.  I missed a line.

This is the fix.

URGENT- because it breaks live right now due to a recent upgrade I did for Scratch to fix a different bug.


BenignCremator:
Looks good, merging.

PR #926, opened 2016-12-12
GBE-710: Fix contact_volunteers

bethlakshmi:
It now gives:
- the list of all volunteers who applied - to this conference
- the list of any events the volunteers have been set up for where the person is a volunteer and the event is in this conference

And - when a profile is added as a volunteer, a volunteer application (a bid) is created to mark that person as a volunteer.  If there is a bid already, it’s marked as submitted and accepted.

Added more tests to cover the challenging logic on this.  

Tuned some of the existing tests that weren’t mocked so specifically.


BenignCremator:
py.test and code review look good.  Moving on to functional test.

BenignCremator:
Right...  Do not test with GBE Admin account, it does not have a complete profile, which generates errors....  Do we want to either make a complete profile for gbeadmin, or alter volunteer review to 'fix' the problem?  A redirect to completing a profile before filing out the volunteer form is also an idea.

BenignCremator:
Seems to work.  DL'd contacts, then drafted a volunteer that was not in the contacts, re-DL's the contacts, then there she was.

PR #928, opened 2016-12-14
GBE 679 cleanup #3:  move people from gbe.models.remains

bethlakshmi:
move all the classes around people

But none of the bids (staying away from dreaded Volunteer as long as possible).

- Existing pepdiff problems remain in the profiles.py file that are carry over from remains.py.  

Crossing fingers that this works on Hunter’s system


BenignCremator:
Works, tests ok, functionality seems alright.  Approving and merging.

PR #934, opened 2016-12-18
GBE-679 cleanup #4: Move out TechInfo

bethlakshmi:
I went small here - it’s moving the Tech Info collection out of the remains.

After this, all that is left is the bidding collection - which is closest to our problem area.
 
Hoping to get this in before @BenignCremator gets into the ticket for audio & lighting views.


gawaineo:
Ran py.test and test results look good. Looked at the code and everything seems in order. Browsed through the website and everything looks good. I approve!

bethlakshmi:
on most stuff we've now been doing single reviewer, since there's only three of us... but on the refactoring in this area, i'd definitely like to wait for @BenignCremator since he's got the machine that doesn't like things. :)

bethlakshmi:
on most stuff we've now been doing single reviewer, since there's only three of us... but on the refactoring in this area, i'd definitely like to wait for @BenignCremator since he's got the machine that doesn't like things. :)

BenignCremator:
Looks good, tested out ok.  Code read well.  Functionality seems to be there.  Approving and merging.

PR #936, opened 2016-12-21
GBE 679 cleanup #5: inch by inch

bethlakshmi:
I moved the simplest stuff I could find out of remains.py - this is the stuff that is Biddable, but only lightly connected elsewhere.


BenignCremator:
Ok, looks good.  Setup vbox and ran pytests and runserver.  No errors.  Code review looked good.

PR #937, opened 2016-12-22
GBE 916: email on any bid state change

bethlakshmi:
OK, I found a pretty great email system - django-post-office:

https://github.com/ui/django-post_office

It does what I wanted with templates and also tracks the emails and makes them accessible in /admin.

To Begin manual testing
———————————
Manually reprovision vagrant
install/create the DB of your choice
./manage.py migrate
in local_settings.py, install:

EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
DEFAULT_FROM_EMAIL = 'test@localhost.com'

——————
Then accept or deny or otherwise change of any bid, and the email will;
- show up in console if sent with priority=now
- show up in the post office section of /admin
- note - it gives a plaintext and an email
- note - an email template is now created and editable by user.

Work included: 
- refactor change-state to use inheritance to avoid code duplication
- change all mail sending to use post office
- more test coverage
- found a mail send bug in sending two emails to volunteer reviewers (old code)
- pulled mailing out of models to make a consistent layer for email distribution (gbe.functions)


bethlakshmi:
I'm doing a test install of this on our test server to iron out the kinks prior to merge...
Notes for a webfactional install:

1.  Do "pip install --user django-post_office"
2. Go to local_settings.py and add "EMAIL_BACKEND = 'post_office.EmailBackend' "

?problems with cron jobs have me taking a pause.

BenignCremator:
Tests and code review look good, checking functionality.

BenignCremator:
I like it.  Worked well, and better to use standard modules for something like this then custom code.  One bit of weirdness: a template only appears in admin/post-office/templates after the first time it is used.  Before that, they are not listed.  ???  Odd, but it seems to work.  Approving and merging.

Issue #896, opened 2016-10-01
Research coordinating ticket prices/descriptions


burlexpo:
We want to have ticket descriptions and prices automatically synched.  When synched, information from BPT takes precedence over information on Burlesque-Expo.com

Related to #238

bethlakshmi:
Putting this down because I think I can the other unassigned & open ticket done more quickly.

bethlakshmi:
Picking this up, but I hit a snag.

The description is easy enough to find, and it's easy enough to always update ticket info when the description changes.

The challenge is coordinating the prices...

The BPT API does NOT give me a clear indication on what price to publish.  What it gives me:

http://www.brownpapertickets.com/apidocs/2.0/pricelist.html

is the field "live":

- "yes" = currently available for purchase
- "no" = currently NOT available for purchase

BUT - that includes things that can be purchased with coupons.  So if you use our developer id, and poll this api:

http://www.brownpapertickets.com/api2/pricelist?id=**PUT_DEVELOPER_ID_HERE**&event_id=2504574&date_id=1408005

Then you actually get 3 prices that are "live":
- $175 - the main purchase price right now
- $40 - a deal that can only be bought with a coupon by very special people
- $0 - a deal that can only be bought with a coupon by VERY very special people

We only want to put the first on our site.

So... I can't *totally* trust the value of "live".

I have two options:

1 - we argue with BPT and get them to add some sort of coupon indicator to the API. That would tell us if the price is fully public, or something we only offer to coupon holders.

2 - I add a "coupon_required" field to our side of the data.  If you set "coupon_required" - the code will never ever publish that price.  If you *don't*, then when the price goes live, we change the public price.

Also - to make this really useful, we need a timed sync with BPT to get the ticket prices on a regular basis, so we can automatically change the price when needed.  In most cases, the conditions are that we sold X number of cheaper tickets, and that trigger is unpredictable.

PR #942, opened 2017-01-01
GBE-896: BPT auto-update ticket description and publication

bethlakshmi:
See #896:

**Thing 1:**
Because of how BPT works, I added two new fields:
 - live - is BPT currently publishing this price?
 - has_coupon - is this a private/coupon driven price that should not be made public?

The two together now define what used to be ‘active’ - the state of whether or not the ticket price is visible.

Also - when we sync, the description, live, and cost fields are pushed from BPT instead of remaining true to the expo data.

**Thing 2:**
added syncing for tickets to the cron job code

**Thing 3:**
Refactored the listings for ticketing associated with each event detail to make sure that only the most expensive active ticket is listed (in case of user error or sync issues)

In so doing, I pulled the logic out of models into a ticketing function.

**Thing 4:** 
Fixed the list for ticket buying for all options to be the same logic as thing 3 - but for all events on all upcoming or current conferences.

The changes for thing 3 & 4 were easier to do as a dict, so I refactored the templates and pulled business logic from the templates, as well.

**Summary:**

Also wrote a bunch more test cases.
And refactored one of the tests to be the style we currently use.
Unfortunately, I added two lines of untested code, because I couldn't figure out how to test the cron job code.

This was definitely a bigger refactoring job than I planned.  Sorry about that, but the logic in ticketing had been spread pretty far and wide.  Hopefully it's more consolidated now.


bethlakshmi:
Please resume testing - I fixed the circular import on transaction_cron_job.
It's now installed and working propertly on webfactional test.

BenignCremator:
Bethlakshmi put the code up on gbetest, and I tested it there.  Seems to work.  Double checking with her if there is anything specific to check, otherwise looks good and ready to approve.

BenignCremator:
No, nothing more.  Approving and merging.

Issue #856, opened 2016-08-13
BUG: All video options displayed when none selected


burlexpo:
When submitting an act, a performer can select one of 
- This is video of the act I'm submitting
- This is video of me, but not the act I'm submitting
- I don't have any video of myself

Ordinarily in reviewing an act, only one of those options is displayed.  In looking at [Red Velvet's submission ](http://burlesque-expo.com/act/review/715) where (my guess is) she didn't select any of those options, we sell _all_ of those options displayed. 

![red velvet review](https://cloud.githubusercontent.com/assets/5369404/17641153/30e5fae6-60e2-11e6-8e87-bc94af7e5d45.jpg)

bethlakshmi:
now up on test.

burlexpo:
Tested.  Works.  Closing.

PR #943, opened 2017-01-01
GBE-856: show only video selection

bethlakshmi:
addresses #856 using the same trick used on other tickets - overload the form with only the choice that was selected.

Since the bug affects both ‘view’ and ‘review’ - I refactored them both to use a shared function.


BenignCremator:
Pytest and functional test both looked good.  Reviewing code now.

BenignCremator:
Code looks solid on read-through.  One style note, however.  If an import is only importing a single item, it is better to not put it in brackets, and keep the import on a single line.  See act_display_functions.py, importing from django.forms.  Not a blocker, approving and merging.

PR #944, opened 2017-01-02
GBE 546: remove hard coded dates

bethlakshmi:
All the hard coded dates are dead code.  I removed them and refactored the one set that was actively in use to the file that was actually using it.

Reran all tests
gripped for any cases of the code - couldn’t find any


BenignCremator:
Code review looks very good.  Nice deletions.  Functional test also good, things seem to work.  Py.test is happy, no errors.  Approving and merging.

PR #951, opened 2017-01-06
GBE 950: separate conference list from schedule build

bethlakshmi:
I found a pretty clean, pretty small point of separation in this one function.

it not longer knows anything about gbe.models - but this increases the need to import gbe.model elements in gbe.functions — > many of which were originally failing in the original problem.

### DIFF
--- a/expo/gbe/functions.py
+++ b/expo/gbe/functions.py
@@ -227,3 +227,28 @@ def send_schedule_update_mail(participant_type, profile):
             'profile': profile},
         priority='now',
     )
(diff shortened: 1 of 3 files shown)


BenignCremator:
Tests were great.  Code works (which considering that it cleared a bug that prevented the server from running at all, major improvement), runcoverage was clean.  Code review looked good, pulled some things that should have been in gbe out of the scheduler, so big win for clearer and cleaner code.  Approving and merging.

PR #952, opened 2017-01-07
GBE-664:  Volunteers can Edit bids

bethlakshmi:
This was big due to the #664 functional spec… the work was:
- extend the edit capability to volunteers
- do a bunch of alerting if the volunteer edits the schedule to cause a schedule problem (see notes on ticket)
- remove the volunteer from any volunteer assignment conflicts

But that also means I needed to change the landing page template (to show ‘edit’)

And create (to prevent more than one volunteer option for a given conference.

While I was at it, I refactored the bidding a bit because i could see a way to share code more easily by doing so.  And I found django.sites, which let me avoid hardcoding.

No migrate required.


BenignCremator:
Rucoverage and functional tests both looks good.  Reviewing code now.

BenignCremator:
Code looks good.  Approving and merging.

Issue #961, opened 2017-01-10
Send unregistered Volunteers to ../accounts/register


burlexpo:
Currently, if someone doesn't have an account and tries to volunteer, they're sent to our rather unwelcoming welcome page.

If someone tries to Volunteer and they are not logged in, please redirect them to http://burlesque-expo.com/accounts/register instead.

Also, edit the text on http://burlesque-expo.com/accounts/register so the first sentence reads "In order to do most other things at The Expo, we need you to register for **a free** account".  Bold is only added for emphasis here; don't make the text bold when it displays on the page.

PR #962, opened 2017-01-10
GBE 961: make volunteer registration easier

bethlakshmi:
#961 

Fix the flow to go to registering an account on the premise that most volunteers are new to the expo.  As it turned out “login_required” as an annotation will ALWAYS route you to the login page, not to the account page.  So I had to remove it.  Our logic inside the function protects us anyway.

Also fix wording.

Since I tweaked the logic, I also updated the test to be more specific (it would have passed either way).

### DIFF
--- a/expo/gbe/templates/gbe/register.tmpl
+++ b/expo/gbe/templates/gbe/register.tmpl
@@ -8,7 +8,7 @@
 <H3>Create an Account</H3>
 
 In order to do most other things at The Expo, we need you to register
-for an account.  If you're planning on applying to perform, want to
+for a free account.  If you're planning on applying to perform, want to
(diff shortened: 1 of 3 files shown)


BenignCremator:
Pytest looks good (*).  Not much in code review, both lines look good.  Moving on to functional testing.

BenignCremator:
Ok, functional tests look good, does what it is supposed to.  Approving and merging.

That said, in my mind, having a page with both login and account reg blocks would make sense, since we do not know if someone hitting the volunteer button either does not have an account, or simply has not logged into their account yet.  Looking at the webpage on my restricted laptop screen makes it seems like there is room to make a nice layout with both.  Something to consider in March.

PR #965, opened 2017-01-12
GBE 945&709: show accepted class, and add a way to put an image for teachers with no photo

bethlakshmi:
To manually test - 
- 709 - if you’re using recent data, compare: 

http://www.burlesque-expo.com/bios/teachers?conference=GBE2016

and 

localhost (same URL)

Around ‘Roman Sirotin’ the view is messed up on the site, and OK on the  fixed branch.  Also - if you want to check the django-cms - you can now add a placeholder image via django-cms, so that I don’t have to care what image Scratch wants.

————

945 - also notice that if you use the current conference (http://localhost:8282/bios/teachers), the accepted classes show up, even if they aren’t scheduled yet.

Unscheduled classes don’t have details that show up in the scheduler views, so the teacher’s classes are listed by title w no link when they are unscheduled, and with a  link after they are scheduled for a time and date.

### DIFF
--- a/expo/gbe/templates/gbe/bio_list.tmpl
+++ b/expo/gbe/templates/gbe/bio_list.tmpl
@@ -1,5 +1,6 @@
 {% extends "base.tmpl" %}
 {% load ticketing_filters %}
+{% load staticfiles cms_tags %}
 {% block title %}
   Great Burlesque Exposition: {{title}}
 {% endblock %}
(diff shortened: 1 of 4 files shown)


BenignCremator:
pytest and code review look good.  Moving on to functional test.

BenignCremator:
Functional test looks good.  Formatting is correct.  Approving and merging.

Issue #699, opened 2016-01-30
Volunteers must have profiles with Badge Names


burlexpo:
Create some sort of check that there is a name associated with a volunteer bid.  If there isn't one, put in their badge name.  If they don't have a badge name, put in their first name.  If they don't have a first name (i.e., the don't have a profile) keep redirecting them to the profile page until they get it.

jonkiparsky:
There are a few things to get straight here. 
First of all, a user's first name comes from their User object, not from their profile. There is one user in the system without a first_name, and that is the "limbo" user that is used in the ticketing system. (I don't remember the details of this, it was some sort of sketchy Marcus code, but anyway there's a magic user in the system that is used by the ticketing module) All other users have first_names.

Second, all Volunteer bids are associated with profiles. You can't have a volunteer bid without a profile, it's a database constraint. 

Third, all volunteer bids currently have profiles with display_names. (just confirmed this about two minutes ago)

So what exactly is the problem you're trying to solve here? Give me the symptoms.

burlexpo:
> So what exactly is the problem you're trying to solve here? Give me the symptoms.

The symptoms are volunteers who do not have names displayed when they come through.  Yes, all of the volunteer bids currently have profiles with display_names.  That's because I've fixed the ones that have come through without names by hand as they've come through.

There is a way to circumvent the DB constraint of "you can't have a volunteer bid without a profile" quite accidentally and enthusiastically.  It's easier to show you in person -- and since I've got a work-around for this (and there's almost certainly going to be more pressing stuff for you to work on), I'm going to drop this down to 'after the Expo'.

bethlakshmi:
All of the users showing @burlexpo's systems are NOT coming through ticketing (the source of the limbo user, who is exactly 1 user and profile in the system to accommodate purchasers in BPT who never logged into the expo site).  

In all of these cases, the user had a legitimate username, firstname, lastname and password (all the stuff in the User object)

HOWEVER - they also had a near empty Profile object.  No badge name, and other fields that we require on the update profile page are empty.  This can happen through are regular login by the following steps:
1 - register a new account - http://www.burlesque-expo.com/accounts/register
2 - get redirected to the profile page (/update_profile)
3 - choose not to finish it, go to the I want to... Volunteer! link and proceed through the volunteer bid process.

You will end up with a user object, (first name + last name), and a Profile Object with no display name, no phone (also required), and therefore, no name in many other parts of the system.

[...]

PR #968, opened 2017-01-15
GBE 699: prevent volunteers from submitting with no phone or badgename

bethlakshmi:
There are details on the bottom of #699 of all the navigational paths.  I made sure that no matter what you do, the system will detect what you’re missing and get you to the right screen to fix it.

Put in the error message we wanted when a volunteer is found to be missing profile info.

Put in check that a profile is not “complete” if it does not have a display name and an email.

### DIFF
--- a/expo/gbe/models/act.py
+++ b/expo/gbe/models/act.py
@@ -19,6 +19,7 @@ from gbe.models import (
 )
 from gbetext import (
     acceptance_states,
+    act_alerts,
     act_not_unique,
     video_options,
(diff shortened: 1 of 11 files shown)


bethlakshmi:
Fixing merge errors...

BenignCremator:
Pytest and functionality looks good.  Reading the code now.

BenignCremator:
Ok, code review looks pretty good.  Approving and merging.

PR #969, opened 2017-01-15
GBE 912: only show the shows for this conference

bethlakshmi:
The manual test is pretty easy:
- use a recent dump of GBE
- go to ‘http://localhost:8282/scheduler/acts'
- select a show
- see that only 4 shows and “—“ show up.  And they load quite fast.

That’s it.  If you look at master, you’ll see the list gruesomely long, loads slow and expands off the real estate.


BenignCremator:
Pytest and functional testing both look good.  Going on to read code.

BenignCremator:
Code review looks good, approving and merging.

PR #972, opened 2017-01-15
GBE 947:  Better Admin

bethlakshmi:
This came from needing to do a big weird export, but I found a lot of places where we could update admin for easier data manipulation/visibility.

But more tuning, meant more untested code, so I wrote Admin unit tests to make sure we don’t break anything with refactoring.


bethlakshmi:
leaving the "ready for review" tag after pulling to signify that this was NOT reviewed by anyone but me.

Issue #957, opened 2017-01-08
Shows not displayed on ../act/reviewlist


burlexpo:
If I go to http://burlesque-expo.com/act/reviewlist all information is displayed where it should be except the column headed "Show" is blank.  i.e., If someone has been Accepted for The Main Event, the "State" column shows 'Accepted', but there's nothing in the 'Show' column.

PR #973, opened 2017-01-15
GBE 957: Show shows on act/reviewlist

bethlakshmi:
#957 is a bug that came from refactoring the models.py into /models/*.py

In so doing, we didn’t import ResourceAllocatino, and the way the code was written, it failed silently and had no unit test backup.

Refactored and added code to:
- use scheduler in a way that means less cross-model contact.
- add unit tests to keep failure from remaining silent.


gawaineo:
Ran pytest, checked out the changes visually...everything looks good, I approve!

PR #974, opened 2017-01-16
GBE-970: sortable, consolidated /profile/manage

bethlakshmi:
Merged 3 fields into contact info.
Made the sortable JS-powered table.
Pulled display info out of the model
Made missing unit tests.

The pepdiff errors on profiles.py come from the stuff that’s always been there.


gawaineo:
Ran py.test, all tests passed, tested sorting on the profile/manage page for users. The result looks great! Approved!

PR #977, opened 2017-01-19
GBE 949: Accept Act is now show-sensitive

bethlakshmi:
Change = when an act bid is accepted (Review -> Acts, select act, click Review, change the form on the right to accept and submit) - then the “act accepted” email now tells them what show they were cast into and creates/uses a specific template for that show (so there is 1 template for rhinestone review, 1 template for main event, etc).

Waitlist, reject, withdrawn, etc - do not do a per show view.

Fun fact:  there are 525 lines out of 6815 of code that are NOT tested.  A month ago, there were ~583 lines.  If we hit 504 lines of untested code, we will be at 93% coverage.

### DIFF
--- a/expo/gbe/functions.py
+++ b/expo/gbe/functions.py
@@ -185,7 +185,7 @@ def send_bid_state_change_mail(
         status,
         show=None):
     site = Site.objects.get_current()
-    context={
+    context = {
         'name': badge_name,
         'bid_type': bid_type,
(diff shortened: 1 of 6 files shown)


BenignCremator:
It all looks good.  Read code, Show was in email (found it in admin/post-office/emails), and tests were all good.  Approving and merging.

PR #980, opened 2017-01-21
GBE 979: Speed up assign volunteers

bethlakshmi:
On my test set, this brought me down to a little worse than 50% faster and with 1/2 the SQL queries.

To be even faster, I put caching back on the page.

Added django debug tool bar back, and had to fix some dependancies - if you want to test this with django debug toolbar, you’ll probably need to retrovision your vagrant.


BenignCremator:
Ok, functionality is faster.  Tests look good.  Code review is good.  Approving and merging.

PR #984, opened 2017-01-27
GBE 983: display by half hour

bethlakshmi:
This was all visual fixing, so no unit tests were added.  Same level of coverage as master.

Changes:
- changed the size of a block (i.e., a row in the table) to be 30 minutes
- this made the far left column ugly, so I tuned it to only show time, and have a fixed width
- to compensate for removing the date - I put the date at the top as a header.  Given that the function does both per-day, and a date range, I offered both options for date display.

- pepdiff = no problems
- unit tests - all run
- run coverage = 528 missing lines

### DIFF
--- a/expo/gbe/static/styles/base.css
+++ b/expo/gbe/static/styles/base.css
@@ -842,3 +842,8 @@ span.opps-errors > ul
 {
   padding: 0px;
 }
+
+.calendar_table .cal_body TH
+{
+  width: 80px;
+}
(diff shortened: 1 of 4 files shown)


gawaineo:
Hi @bethlakshmi ,

Which pages can I visit in GBE to see the tables showing the time affected?

gawaineo:
Py.test results good, performed a visual check. everything looks great I approve!
30_min_scheduler

PR #989, opened 2017-02-05
GBE 683: report for audio + more

bethlakshmi:
- refactored reporting into a single directory
- added view_techinfo which creates 4 different views of the act tech info - all, stage_mgmt, audio, lighting,
- similar to review_act_tech_info in most ways - selectable conferences, shows, report formats, and user can export the data from anywhere.

This has been waiting to be done since the holidays - please review by Wed.


gawaineo:
Ran py.test and pepdiff, everything looks good.

One issue I see is that the for the view_techinfo, information is stretching off the page as you can see in the picture below. Not sure how much of a blocker this is.
view_techinfo_stretching

bethlakshmi:
I"m going to decide that that one will take a future ticket.  Putting all this info on a page was something @dracus wanted as part of the bigger picture work.  There's just no way to scrunch it all into a vertical layout this narrow without doing serious design. :(  So when we have a clue on that design, let's write a ticket.

bethlakshmi:
I"m going to decide that that one will take a future ticket.  Putting all this info on a page was something @dracus wanted as part of the bigger picture work.  There's just no way to scrunch it all into a vertical layout this narrow without doing serious design. :(  So when we have a clue on that design, let's write a ticket.

Issue #683, opened 2016-01-29
Create an "Audio Report" page that can be viewed on-line


eclecticmagpie:
**Task List**

Create a report accessed through _Special > Reports_ which reuses the Conference Selector and Show Selector logic to access a view which displays all audio related data from _acttechinfo_ in columns.  
Should have JavaScript implemented to allow user to reorder or hide columns.

----------------------------------

I thought I had put in this request last year, but I don't see it, so I'm putting it in now.

The work-around I used last year is broken; I'll list that as a separate issue with a higher priority.

If I were LD/TD, I would also want a view that had similar items from the Lighting Needs section.  However, the view could easily become unwieldy with all that stuff in a single view, so I would suggest having an "Audio Tech" view and a "Lighting Tech view".  Even better would be to have a custom view where whoever is looking at it can add or delete columns of interest to the view, and save those preferences locally via a cookie, or better still to their account info so that they see their preferred view whenever they log in from any machine.

burlexpo:
The summary is so the SM can quickly see who has scheduled a rehearsal time and uploaded their music.  The info you want is all in the act tech download and you can hide or reveal whichever columns are applicable.

eclecticmagpie:
You have not closed this request.  You have blown it off.  That's neither
polite nor appropriate.  If there isn't time to do it this year, that's
fine -- that's why I listed it as a medium priority rather than high.  But
it remains open.

The "download & edit" approach is adequate for a runtime doc, but awkward
for a quick check to see what is being requested so I can design the sound
system and source any necessary gear, or notify you and/or the performer of
our inability to source such gear, in a timely fashion.

Since I start wanted to look at this info well before the show and changes
keep coming until the day of the show itself, doing a daily download & edit
rapidly gets to be a PITA, which encourages putting it off until later in
the cycle instead of staying on top of things.

Note that every time there's a change, the downloaded and edited file
becomes obsolete -- there's no (easy, understood) way to update it.
There's also no way to tell when there's *been* a change, except to
download and create ANOTHER COPY so as to be able to compare the two.  The
only _easy_ thing to do is to deny all special requests.





Mike Bergman
www.prospecthillforge.com
www.vintagedancers.org/teadance
www.isebastiani.com
reply-to: eclectic@mit.edu

burlexpo:
> You have not closed this request.  You have blown it off.  That's neither polite nor appropriate. 

[...]

PR #992, opened 2017-02-05
GBE-990:  Fix unicode display problems

bethlakshmi:
Found some new issues from this year’s data - cleaned them up.

This fix will need massaging when we deal with #683 - chances are, we need to do a manual merge on one or the other. 

Also urgent - this blocks stuff we need at expo run time.


gawaineo:
Ran py.test and pepdiff, looks good all test cases passed. Tested out fields in Act Tech Info such as track title and track artist, everything seems to be in working order. I approve.

PR #994, opened 2017-02-08
GBE-978: Fixes the two menu open problem

bethlakshmi:
You can now click on a submenu, then click on a different submenu and
the old one will close, and the new one will open.

Sadly, this does NOT fix the mobile phone  menu problem.  What I did to customize the CSS way back when plays out VERY badly on smart devices.  

Because this offers a meaningful improvement, I’m pulling.  I intend to continue to fight the mobile  battle later.

### DIFF
--- a/expo/gbe/static/styles/gbe_bootstrap.min.css
+++ b/expo/gbe/static/styles/gbe_bootstrap.min.css
@@ -43,6 +43,12 @@ a:active,a:hover{
 	outline:0
 }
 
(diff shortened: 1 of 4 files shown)


gawaineo:
Ran py.test and all tests passed. Tested changes manually by clicking on the "Special" tab and "Show & Events" tab. Results look great, there was no overlapping sub-menus (I compared the local running app to the main website and I can see the difference). I approve!

Issue #986, opened 2017-02-02
Make "Tell Us About Yourself" a required field


burlexpo:
On http://burlesque-expo.com/volunteer/bid/ make the text question "Tell Us About Yourself" required.

If we can do this ASAP and get it to live, that'd be great.  If we can't get it to live by Saturday, 2/4, it's probably not worth it for this year.

PR #1009, opened 2017-04-14
GBE-986: Make Tell Us About Yourself Required

bethlakshmi:
#986 

- update model 
- autocreation of volunteers needed no tweaking.
- add unit test
- has a migration

Be aware - we have two active automation requiring pull requests.


bethlakshmi:
Had to clean up migrations.  Retested with run server manually.

Issue #1004, opened 2017-04-11
Check for Deactivation through out system.


bethlakshmi:
We never really paid much attention to this.  Since I'm about to work on or finish two cases that render users inactive - we should make sure that our system logic treats them appropriately.  This is the list I can think of of places that should be updated to accomodate or avoid deactivated users:

- show users highlighted in red when they are deactivated but listed in:
     - volunteer schedule reports (as volunteer or staff) - http://www.burlesque-expo.com/reports/staff_area
     - bid review pages (everything in Special -> Review)
     - scheduler management pages (as in anything that goes to privileged users and NOT public!)
          - there may not be anything - needs digging

- remove entirely from:
     - any scheduler list in event scheduling
          - http://www.burlesque-expo.com/scheduler/create/GenericEvent/796
          - test both a regular event and a volunteer opportunity
     - reports for printing schedules (schedule info, envelope stuffing instructions)
          - http://www.burlesque-expo.com/reports/schedule/all
          - http://www.burlesque-expo.com/reports/stuffing/gbe2011_summer
          - For Badge Print Run - keep it.
     - user facing group lists - i.e., troupes.  Ideally, a deactivated user will stay in the selected list, but not be newly selectable.
     - email sets in Special -> Contact

Manage --> Users already covers this - deactivated users won't show up there as of #971 
Login already works as part of the default Django user management.

PR #1010, opened 2017-04-16
GBE 1004: obey deactivation in privileged areas

bethlakshmi:
The laundry list is listed in #1004 .

Also had to refactor a few tests that were painfully archaic.


bethlakshmi:
Fixed a conflict - needs retest and pepdiff before merge

Issue #956, opened 2017-01-07
Add a Captcha to ../accounts/register


burlexpo:
Upon reviewing the existing users, I noticed a bunch that were clearly created by robots.  I don't think they're harmful, but they do clutter up the DB and could potentially be harmful.  Can we install something like https://github.com/kbytesys/django-recaptcha2 to validate the forms submitted through ../accounts/register?

The first couple of Captcha packages I looked at on https://djangopackages.org/grids/g/captcha/ seem to want at least v1.7 of Django.  Also, I'm a big fan of the "I'm Not a Robot" verification (https://nakedsecurity.sophos.com/2014/12/05/i-am-not-a-robot-google-swaps-text-captchas-for-quivery-mouse-clicks/) vs the "please type in this distorted text" verification.

gawaineo:
I resource of resources I found on stackoverflow: http://stackoverflow.com/questions/2275806/easy-to-use-django-captcha-or-registration-app-with-captcha

bethlakshmi:
https://github.com/mbi/django-simple-captcha/tree/v0.4.6

bethlakshmi:
How about this one?

https://github.com/praekelt/django-recaptcha

Looks like it uses the checkbox style captcha, and looks easy enough to integrate.  

I also suggest that I put it on any no-login-required form - including class idea proposals.

PR #1011, opened 2017-04-18
GBE-956: Add Google2 Recaptcha

bethlakshmi:
Adding the “I am not a robot” checkbox to the site.  Used this django plugin:

https://github.com/kbytesys/django-recaptcha2

Instructions to test:
- rebuild vagrant (aka, install the plugin)
- set this in your local host:

RECAPTCHA_PRIVATE_KEY = '6Le0dx0UAAAAACNZynxCx5mUovu3M1Au3XFeeFKN'
RECAPTCHA_PUBLIC_KEY = '6Le0dx0UAAAAAFGd_HJzX22FdzhwI-GCh8nCoXoU'

Your machine should read as 127.0.0.1 for this to work.  

These are our test keys and will ONLY work on your local…  the other keys are installed only on the servers.

Since this is a potentially annoying package install, I won’t pull it until I test this branch on the test site.  If there’s no comments, pulling to master and pushing to live will happen quickly after that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,5 @@ expo/gbe/static/mptt/draggable-admin.js
 expo/gbe/static/preview_images/admin-style.psd
 
 expo/gbe/static/sortedm2m/selector-search.gif
+
+random_docs/~$heduling API.docx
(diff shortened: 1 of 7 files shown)


bethlakshmi:
#956

PR #1024, opened 2017-05-06
GBE 888-A: refactor class and act

bethlakshmi:
This is only the first two - doing this so I can work on new Act stuff in the new architecture and not try to perpetuate the old way.


bethlakshmi:
obsolete, better pull coming.

PR #1026, opened 2017-05-06
GBE 889&888:  View classes for create & edit

bethlakshmi:
There’s overlapping code both between bid types and between create and edit.

Combined both flows and all bid types into a super class, and refactored acts and classes.

- added tests for 100% coverage
- updated tests for new state (class names, text changes)
- removed obsolete code from act and class models - this stuff should never be at the model level, the view-based class is a better place for it.


bethlakshmi:
It's incestuous to manually review a pull this large that you made yourself... but I did my best on this.
- re-pepdiff
- cleaned out all *.pyc and reran the tests - that actually found some problems, which I fixed.
- did a manual sanity check.

PR #1046, opened 2017-06-06
GBE-1045: remove fixed width

bethlakshmi:
The fixed width is evil, this got us most of the way, and I fear a more dramatic approach would be something like adding a new plugin and converting code.  Not a bad idea but let’s not let the best be the enemy of the good.


bethlakshmi:
demo'ed to scratch - he's a fan. This is actually the first of two, because once we aren't using old columns or the old carousel, I will kill duplicates and leave us with bootstrap.

Issue #393, opened 2015-03-13
Extra types of Roles for an Event


bethlakshmi:
MC
Featured Guest

Show up special in a Show - not in the grid with everyone else, before it with a snazzy display.

Have a way for the power user to make this stuff show up “Special”

bethlakshmi:
@burlexpo  - how does this look for what you had in mind for selecting featured acts & hosts?

![screen shot 2017-06-27 at 4 37 50 pm](https://user-images.githubusercontent.com/4542707/27608894-13863072-5b57-11e7-9b4b-223ed34eb448.png)

The default is "Regular Act" which shows up in the grid we currently have.

The words shown here for featured performer/host will be what I use when annotating the special stuff.  Anything not a "Regular Act" gets treated as special.  The special list is pretty easily extensible, but it will take a code update.

Let me know:
- that you like the look of it
- any changes to the words in the drop down
- any changes to the text label - "Special Display Selections:" was a real mouthful and makes the screen kind of ugly (OK, this is one wart among many...)
- I'm alphabetizing by performer name in the featured performer display, right?

I'm not blocked, I'll be another few days at the rest of it.

burlexpo:
The mechanism looks fine.  Are the "Special Display Selections" stored somewhere on /admin so I could go and add another one without bugging you?

Do the words in the drop down also appear on the show page? If so, I'd want to change them to "Featuring..." and "Hosted By...".  We could add a third category of "Returning Winners" for use with the winter conferences.

I think changing the label to "Section" is fine and clear.

Yes, you're alphabetizing by performer name within any display group.

bethlakshmi:
OK, first prototype done, and now I really DO need feedback:

![screen shot 2017-06-27 at 11 13 56 pm](https://user-images.githubusercontent.com/4542707/27619119-624b3298-5b8e-11e7-85d7-0f66442a83d2.png)

Here's the image - yes, I faked in a bunch of data to give myself something real to play with.

Problems I see 
**1 - the irregular size of boxes makes this look choppy.  I know I can do the following:**
- set a standard height and force them all to it.  
- (probably w/in my grasp) fix the text to go to a "..." when the last line is reached.

**2 - the irregular image sizes makes a more awkward display** - I have a ticket to force people to do our bidding, we haven't prioritized it - #1052 

**3 - the "featuring" rather blends into the background**, so does the "Check out our fabulous Performers!" which gets smooshed in below (and off the bottom of this page). 
[...]

PR #1066, opened 2017-07-01
GBE 393: make feature acts

bethlakshmi:
- Scratch can define what types of roles a given act fills, le host, feature act, etc. with an admin switch
- act coordinator can “cast” the act into one of these roles in the act review pages.
- the act review table shows the casting as a nice FYI.
- the featured acts of any special type get their own grid display in the event detail view.

This requires a migrate in gbe & scheduler

To use, it has to be manually seeded with feature acts.

#393


bethlakshmi:
re reviewed
pepdiff
all tests pass

Issue #1056, opened 2017-06-22
Classes and other events not mobile friendly


bethlakshmi:
I just tested classes on my phone, it's ugly - the image does not rescale and it gets squashed off the page.

PR #1067, opened 2017-07-01
GBE-1056: Fix look of non-shows in event details

bethlakshmi:
This is all templates and CSS

#1056 

There are two things to test:
- a just-accepted class with no schedule details
- a scheduled class

They are built differently, but should render the same.

May need some fussy checking when the other pull request is merged.


bethlakshmi:
Visual test post merge looks good.

PR #1070, opened 2017-07-06
GBE 1052:  Use Django-Filer to make subject aware thumbnails

bethlakshmi:
Moved performer images to django-filer.  

Key changes:
- performers in shows now have editable links for images & text on show display page to make tuning this easier.
- performer file upload looks largely the same.
- files are “owned” by a made-up super user… need to change user’s password after install so it’s not hard coded in migration.
- when users upload image, that image becomes available in the django-filer system, so it can be reused in the CMS.
- django-filer takes on the work of thumbnail making and updating
- need to remove old images after this system is up and going or there will be a LOT of extra space consumption.
- redid thumbnails everywhere I could find:
- teacher bio page
- class page
- show pages
- landing page

I will let this simmer for a few days and maybe write more unit tests if I can.


bethlakshmi:
Unit tests are as good as I care to make them right now.  All code covered, all the cases that seemed worth it are in.  I drew a line at some template logic.

PR #1093, opened 2017-08-04
GBE-1092: Move around links to data tables

bethlakshmi:
Tricky because I am not 100% sure I caught every case where data tables are used.


bethlakshmi:
found another - fix before merging - volunteer lists on editing events (volunteer opportunities)

Issue #1102, opened 2017-08-08
put styles back... or at least 20th century poster


bethlakshmi:
ideally when editing a text box, you can pick it as a style in the editor (ck editor?)

PR #1107, opened 2017-08-15
GBE 1102:  Fix style plugin, and configure CK Editor

bethlakshmi:
1 - fixed the bad setting in in settings.py that has been there since the upgrade.  Now the Style plugin works

2 - added a new setting to let me configure the styles for CK Editor (the text editor) so the same values are available in the Styles pull down. 

The (2) solution is really powerful and we can do more with it, if @burlexpo gives me other CSS styles he’d like to see in here.


bethlakshmi:
#1102

Issue #1055, opened 2017-06-22
Vendor grid needs mobile update


bethlakshmi:
Follow on from #1047 

http://preview.graygrids.com/item/impression-bootstrap-html5-event-and-conference-template/

is a template I bought.

We like the "blog" for the vendor grid.  Need to update the vendor grid to the blog summary format, and then decommission the non-mobile mobile grid.

burlexpo:
@bethlakshmi I played with this and it mostly looks fine.  Links work as intended, layout looks good.  Editing the header text seems to work fine.

Is there a way to force the images/logos so they fit in the entry? I'm particularly talking about "Glorious Pasties" or "Vixens Ahoy".

PR #1129, opened 2017-09-18
GBE 1055:  vendor bid made into bootstrap with filer

bethlakshmi:
- updated display to use the same basic framework as featured performers
- updated both vendors AND featured performers to have a (more) to open the modal when there is long text
- converted image management to the filer system, same as personas
- saw a bug in personas - should only save the model object when commit=True (fixed it)


bethlakshmi:
updated with comments from @burlexpo addressed from #1055 notes

PR #1135, opened 2017-09-27
GBE 1103:  Make a tool to email to bidders from the website

bethlakshmi:
- used the tricks we already have in the system - namely django post office for email, and tiny-mce for HTML composition
- built a new page, accessible to all coordinators from “Special -> Contact -> By Email”
- page lets you select a filter - by conference, bid type and bid state
- you can only email to bids that you are a coordinator for.
- once the filter is selected, you are given the list of recipients and a tool to send them an email.
- if you are an admin (i.e., a super user) you can pick the “From” address.  If not, YOUR address is used, so you can’t spoof another user.
- ~~whenever the mail is sent, all the recipients are BCC’d and the To address is the same as the From address.~~
- mails are sent separately to each recipient so no one sees anyone else's address.
- it does not select any deactivated users.  They are omitted from all filters/emails
- post office keeps a record of whatever is send in the regular place.  It does not record the USER, it records the From/To, etc.

NOTE - I did not put the same safety control on this, if you mail a user from the test system, they will get it.


bethlakshmi:
review looks good, waiting on 1 more runcoverage

Issue #1167, opened 2017-11-10
Make Volunteer Interest searchable and reportable


bethlakshmi:
Based on the conversations with @burlexpo and @eclecticmagpie - we need a way to let people managing subsets of volunteer shifts w/in a show see, edit, and review their shifts, without having to deal with the show to shift structure.  The case in point is Audio - where the shifts are connected to the work of each show, but it's a consistent audio staff that works during the weekend.

Volunteer interest is the easy way, as it already has a 1 to many relationship with the event and it doesn't alter the existing volunteer to show relationship.

So... the work for me here is:
- make the new event management list (/scheduler/manage) provide a filter for Volunteer Interest - it will be a list of checkboxes at the top of the screen that combine with the other filters for whatever the user needs to search.
- create a report that works like - /reports/staff_area/721 - and give a list of each shift, and the volunteer for each staff area.  There will be some overlap here, as things like "registration" are both staff areas and volunteer interests.

bethlakshmi:
I still have to write tests, but I think this is what we were talking about for the first step:

http://gbeadmin.webfactional.com/scheduler/manage

bethlakshmi:
Reports and filters are done, tests are written.  Haven't pushed to test yet, only partial work is up right now.

PR #1168, opened 2017-11-11
GBE 1167:  Make volunteer type a useful concept for coordinators of areas

bethlakshmi:
Based on Mike’s needs - audio is a part of shows, but we want Mike to be able to focus in on just the audio shifts.  We’ll do that with volunteer_type, by:

- Making volunteer type searchable in the manage event lists (the thing that lets you edit and organize), so he can edit and organize.
- Make a report of who’s assigned to shifts of that type.

Note:
- when you search by volunteer type, you will see only volunteer positions, because they are the only things where we can set volunteer type
- I refactored the reports pretty seriously - they needed to use the new scheduler API, they needed to be broken up like our other types of views, and they were ugly.  I fixed those things.
- I removed about 10-15 lines of unused code.  Main focus is to trim away the weirdly merged scheduler vs. gbe codependency.


bethlakshmi:
#1167

PR #1173, opened 2017-11-18
GBE-1095:  Let users show interest

bethlakshmi:
Let users show interest.

This feature reuses our existing system of booking people for events.  The only addition is to the list of roles - there is now an “Interested” role which is meant to indicate people who have shown a casual interest in the event.  

That relationship is managed through a toggle functions - set_favorite - on or off. 

The GUI then provides a star icon in the following places:
- the calendar
- the event details for an event (near the date/time)
- the landing page (/gbe)

When the user has shown interest, the star is filled in, when they haven’t, it’s an outline.  The user can toggle the star on an off as many times as they like.  Wherever they toggle the star, they remain on the page, and the star is updated.

When the star is turned on (i.e., when interest is indicated) the event is then treated as part of the user’s schedule.  It shows up on their /gbe page schedule, and it shows up (grey color) on their printed schedules that they get when they walk in.

That means that there are now two forms of commitment in our system:
- a booked commitment for teachers, panelists, volunteers, performers, etc - which can NOT be altered by the user, it must be altered by a coordinator
- a casual interest, which is managed purely by the user.

Wherever both forms of commitment are shown to the user, the distinction is made clear:
- on the landing page (/gbe) - they get an active star for interests, and a greyed out star for commitments.  The hover text also makes clear that they have to reach out to  coordinator for the commitments.
- on the printed schedule - the commitments are bold, the interests are grey.  Also, to deal with some vagaries in refactoring, I put on the schedule the nature of the user’s commitment (teacher, performer, etc.)   This was because we have some rare cases (Mina) where the user is committed to an event several different ways and I can’t reduplicate that situation easily while still highlighting the difference between commitment levels.
- on the calendar - interests show a filled in star and a pale color, teaching, performing or volunteering shows a greyed out star and a pale color. 

NOTE:  to betty when she deploys - update the URLs for event detail items that are hard coded in menu - there was refactoring that messed with this.


bethlakshmi:
Make the Calendar Detail page function the same as the Main calendar page.  If the person is obligated to the event, they should not be able to be interested/disinterested.

PR #1186, opened 2017-12-05
GBE 1098: interest view for teachers and coordinator

bethlakshmi:
Here’s the interest view for teachers and coordinator.

I’m actually in favor of @burlexpo’s idea of a badge for the # interested next to the teacher’s schedule item (center column of /gbe) but decided without feedback to skip refactoring.

Still need to do the email tool… putting it off for now


bethlakshmi:
I'm picking this back up and converting the teacher view to a badge.  I thought about it more, and it's better there because:
- if we have a substitute, the substitute gets the info.
- a class that runs at 2 different times gets the different time interest that way
- I can extend it to master and dropin classes very easily

PR #1202, opened 2017-12-23
GBE 1143:  make a privilege for staff lead and give it to common reports & tools

bethlakshmi:
Built out a new way of validating for a privilege.  From the permission check perspective, it’s exactly the same as the other roles.

But from the code perspective, it’s checking the affiliation of a lead with an event.

Also - because hijack now offers a new update, I needed to update the urls.py for the new way of hooking in hijack.


bethlakshmi:
Note to self - upgrade hijack when you do this.

expo/tests/gbe/test_review_profiles_view.py
```diff
@@ -42,6 +43,14 @@ def test_all_well(self):
...
         response = self.client.get(self.url)
         self.assertEqual(200, response.status_code)
+        self.assertContains(response, "Manage Users")
+
+    def test_staff_lead(self):
+        context = StaffAreaContext()
+        login_as(context.staff_lead, self)
+        response = self.client.get(self.url)
+        self.assertEqual(200, response.status_code)
+        self.assertContains(response, "Manage Users")
```
bethlakshmi: Let's add one to test that OLD staff leads (past conferences) DON'T get this privilege.

expo/tests/gbe/test_special_menu.py
```diff
@@ -90,3 +92,27 @@ def test_regular_users_get_nothing(self):
...
+                    menu_item['url'],
+                    status_code=200,
+                    msg_prefix='Role %s gets url %s' % (
+                        privilege, menu_item['url']))
+                self.assertContains(
+                    response,
+                    menu_item['title'],
+                    status_code=200,
+                    msg_prefix='Role %s gets title %s' % (
+                        privilege, menu_item['title']))
```
bethlakshmi: Add one to check that leads of past conferences don't get any privs.

bethlakshmi:
wait until I address my own comments.

Issue #1222, opened 2018-01-05
Attendee printouts


bethlakshmi:
The printouts of schedule/stuff they get should make sure that in any given role, you only get the bullet items for that role once- example case, Devastasia

Issue #1221, opened 2018-01-05
Can't run the registratration checkin report on gbelive


bethlakshmi:
It now runs too slowly, I had to run all this stuff off my laptop.

Minimum - I need to be able to batch the report somehow
Better - I need to be able to run it for specific users one by one.  And also run the whole set in a more performance friendly way.

PR #1236, opened 2018-03-08
GBE 1221:  Make welcome letter faster and less wasteful

bethlakshmi:
This report timed out on wdbfactional.

it will need a retest.  Key elements:
- we only print the report if you’ve got a schedule.  Ticket buyers only don’t get the report.
- refactored to use the scheduler roles function, which is more efficient
- put in a user-by-user function, and linked that to the manage profiles page, so that even if this doesn’t work on a live saver, registration can print one offs. 
- when printing for the individual, ANY user will work, even if they have no schedule.


bethlakshmi:
Also #1221 and #1222

Issue #1241, opened 2018-03-27
Post-class eval


burlexpo:
Logged in as myself (Scratch, priv. user) on Chrome.
I go to http://gbeadmin.webfactional.com/reports/evaluation
I look at some class evals.  All is good.
I click the eyeball to get some details and the pop-up comes up.  Also good.
I dismiss the pop-up window.
I go back to the _Special_  menu and pull it down to move to another section.  I get the most recent pop-up I looked it, instead of being able to access the rest of the Special menu (that is, I can still see the top level items on the Special menu -- Contact, etc. -- but any of them that I select and click on gives me the pop-up window.
I can click one of the other menus (for examples, _Shows_) and have it work as it should.

bethlakshmi:
I can reproduce it!  Nice find!

I found another bit of ugliness while testing, will fix both before pulling onto the live site.

PR #1244, opened 2018-04-06
GBE-1241:  Fix JS and layout problems

bethlakshmi:
Had 2 ugly layout problems that I fixed.
And fixed the JS bug that was connected to problem from #1241


bethlakshmi:
make sure to test different layout on web factional

PR #1256, opened 2019-04-20
GBE-1253: add to the event_settings

bethlakshmi:
The blocking error in editing a rehearsal slot came from the event_settings not having an item for a rehearsal slot.


bethlakshmi:
pepdiff and full test passed.

bethlakshmi:
Fixes GBE-1253

Issue #1254, opened 2019-04-20
!@#!@#  CSS !!!  - Act review page


bethlakshmi:
![Screen Shot 2019-04-20 at 2 10 27 PM](https://user-images.githubusercontent.com/4542707/56460800-8f1f6180-6376-11e9-9db0-aa8d6d72a336.png)

Need I say more.

This isn't on production yet and should never go there.

PR #1257, opened 2019-04-20
GBE-1254: Add a class for the act header div

bethlakshmi:
The act review header (not the footer) is the only place where the existing transform/offset is a bad thing.  Since the other 3 work fine (in class evals, and in the footer), I made a special override for the div for the header for rotated header names.

That seems to work visually.
No changed pepdiff
#1254


bethlakshmi:
testing works.

